PR #3, opened 2015-03-08
Added support for Cinder.

kat-co:


cinder/autogenerated_client.go
```diff
@@ -0,0 +1,1821 @@
+package cinder
```
axw, line 12: All source files need a copyright header.

cinder/client.go
```diff
@@ -0,0 +1,264 @@
...
+import (
+	"fmt"
+	"net/http"
+	"net/url"
+	"time"
+)
+
+type TokenFn func() string
+
+func SetEndpointFn(endpoint *url.URL, wrappedHandler RequestHandlerFn) RequestHandlerFn {
```
axw, line 27: Seeing as the Host field is the only part of `endpoint` that is used, just pass in host?

axw, line 27: This should be documented.

cinder/client.go
```diff
@@ -0,0 +1,264 @@
+package cinder
+
+import (
+	"fmt"
+	"net/http"
+	"net/url"
+	"time"
+)
+
+type TokenFn func() string
```
axw, line 23: doc

cinder/client.go
```diff
@@ -0,0 +1,264 @@
...
+
+func SetEndpointFn(endpoint *url.URL, wrappedHandler RequestHandlerFn) RequestHandlerFn {
+	return func(req *http.Request) (*http.Response, error) {
+		req.URL.Host = endpoint.Host
+		req.Host = endpoint.Host
+		return wrappedHandler(req)
+	}
+}
+
+func SetAuthHeaderFn(token TokenFn, wrappedHandler RequestHandlerFn) RequestHandlerFn {
```
axw, line 37: doc

cinder/client.go
```diff
@@ -0,0 +1,264 @@
...
+func NewClient(tenantId string, handleRequest RequestHandlerFn) *Client {
+	return &Client{tenantId, handleRequest}
+}
+
+type Client struct {
+	tenantId      string
+	handleRequest RequestHandlerFn
+}
+
+// Shows information for a specified snapshot.
```
axw: These aren't auto-generated, right? Would prefer to stick with conventional "GetSnapshot ..."

cinder/client.go
```diff
@@ -0,0 +1,264 @@
...
+}
+
+// Lists volume types.
+func (c *Client) GetVolumeTypes() (*GetVolumeTypesResults, error) {
+	return getVolumeTypes(c.handleRequest, GetVolumeTypesParams{TenantId: c.tenantId})
+}
+
+type StatusResultFn func() (string, error)
+
+// VolumeStatusNotifier will check a volume's status to determine
```
axw: doc comment is inconsistent with method name

cinder/client.go
```diff
@@ -0,0 +1,264 @@
...
+// whether it matches the given status. After a check, it waits for
+// "waitDur" before attempting again. If the status does not match
+// after "numAttempts", an error is returned.
+func (c *Client) StatusNotifier(
+	getStatus StatusResultFn,
```
axw: I suspect the two first params would be better as a single predicate that returns (bool, error). That'll allow for alternative  strategies, like waiting for the status to be in a set, or _not_ equal to something, etc.

cinder/autogenerated_client.go
```diff
@@ -0,0 +1,1821 @@
...
+			Capabilities string `json:"capabilities"`
+		} `json:"extra_specs"`
+		ID   string `json:"id"`
+		Name string `json:"name"`
+	} `json:"volume_type"`
+}
+
+//
+// Updates a volume type.
+func updateVolumeType(request RequestHandlerFn, args UpdateVolumeTypeParams) (*UpdateVolumeTypeResults, error) {
```
axw, line 59: I wonder if these shouldn't be exported. We might periodically regenerate these, but not add exported methods on Client for every function. Then, others can make use of these auto-generated functions without having to go through the process of adding an exported method, writing tests, etc.

cinder/autogenerated_client.go
```diff
@@ -0,0 +1,1821 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strings"
+)
+
+type RequestHandlerFn func(*http.Request) (*http.Response, error)
```
axw, line 26: It'd be good to document this, and describe that the auto-generated functions will have a hard-coded URL that needs to be overridden by the `RequestHandlerFn`. Maybe move it out of the auto-generated file as well.

cinder/client.go
```diff
@@ -0,0 +1,264 @@
...
+	desiredStatus string,
+	numAttempts int,
+	waitDur time.Duration,
+) <-chan error {
+	notifierChan := make(chan error)
+	go func() {
+		for attemptNum := 0; attemptNum < numAttempts; attemptNum++ {
```
axw, line 254: `defer close(notifierChan)` prior to the for loop
No need for the `notifierChan <- nil` below then.

axw:
Generally looking good, just a few comments. I'd like to see the code generator checked into go-goose (or a separate repo), and a `//go:generate` directive added to one of the Go files to create `autogenerated_client.go`.

client/client.go
```diff
@@ -146,6 +152,11 @@ func (c *client) MakeServiceURL(serviceType string, parts []string) (string, err
 	return makeURL(c.baseURL, parts), nil
 }
 
+func (c *authenticatingClient) HostForRegion(region string) string {
+	return ""
+	//return c.regionServiceURLs[region]
```
wallyworld: ?

cinder/autogenerated_client.go
wallyworld, line 26: I'd like to see a bunch of doc describing the auto generation process, a link to the generation utility, where the WADL is sourced from etc.

dimitern:
@katco- do you plan on adding a cinder test double package later to support local/live tests needed for juju, like we have for nova and others?

cinder/autogenerated_client.go
wallyworld, line 59: In my view, I'd like to refrain from exporting these at the moment. The user facing, exported functions take params divorced from the params structs used on the wire etc. I think it's good to be able to explicitly control the user facing interface.

kat-co, line 59: @axw: Originally I did have these exported. I agree that it would probably be helpful to give the user flexibility in how they interact with the API.

@wallyworld: These are the same types that are utilized in the public functions, and they will always represent the information the user must specify in order to make a request. I agree that we should offer a curated experience, but I think I disagree that we should try and hide the raw calls from developers. If we make a breaking change, we can always use a version bump.

I am open to this idea. Can you give me an example of where this might be problematic to help clarify?

cinder/client.go
kat-co, line 254: Great suggestion!

cinder/client.go
kat-co: No they are not. Apologies, I think I just copied the documentation from the WADL and forgot to convert to Go format.

cinder/client.go
kat-co: Excellent suggestion!

client/client.go
kat-co: Whoops, development artifact! I've removed this entire function.

cinder/client.go
kat-co: Corrected.

cinder/client.go
kat-co, line 23: Corrected.

cinder/client.go
kat-co, line 27: Documentation corrected.

Regarding the endpoint parameter, I had this same debate with myself. The reason I went with URL is because I wanted a well-formed host, and URL guarantees this. I could take in a string and validate in this function, but that breaks the desired signature by returning an error (this makes it less "nice" to chain wrapper functions).

It's definitely debatable... given that new information, do you have any opinions/suggestions?

cinder/autogenerated_client.go
kat-co, line 12: Corrected.

Just FYI, it looks like no other files have the copyright. We might want a separate PR to correct that.

cinder/client.go
kat-co, line 37: Corrected.

kat-co:
@axw Regarding the auto-generation, and the inclusion of a go:generate directive. I'm going to look into this a bit, but one thing that immediately strikes me as problematic is referencing the external WADL files. I'm not immediately sure how to responsibly do that. Vendoring seems problematic at first glance.

cinder/autogenerated_client.go
axw, line 59: @katco-: If you do go ahead and export them, I think it'd be nice to have the generated and hand-written code in separate packages. Along the lines of the "gen" subpackage in https://github.com/awslabs/aws-sdk-go. Doing that would make it very clear that the functions are auto-generated, as opposed to part of a well-tested, curated set of functions.

axw:
> @axw Regarding the auto-generation, and the inclusion of a go:generate directive. I'm going to look into this a bit, but one thing that immediately strikes me as problematic is referencing the external WADL files. I'm not immediately sure how to responsibly do that. Vendoring seems problematic at first glance.

Is there a URL for the WADL file(s)? The generate command can be arbitrarily complex, so it can grab external files if necessary.

I'm not sure about vendoring. go-aws does do that, from what it looks like. It may also make sense to vendor the version of the WADL files that were used to generate the source files in the repo.

cinder/autogenerated_client.go
kat-co, line 26: I've added a bunch of documentation at the top.

cinder/autogenerated_client.go
```diff
@@ -0,0 +1,1704 @@
+// Copyright 2015 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
```
axw: Eep. Libraries should be using LGPLv3, and the root LICENSE file confirms that.

cinder/autogenerated_client.go
axw, line 12: Oops. Yes.

axw:
With license header fix, LGTM.

cinder/autogenerated_client.go
kat-co, line 59: @wallyworld and I discussed this at one point. I think it's a good idea, and the only reason I can see to keep it in the same package is that we make extensive use of the autogenerated types. I'm not sure if it makes sense to have to reference a 2nd package for that. Any opinion on that?

cinder/autogenerated_client.go
kat-co: Ack! Good catch. Thank you!

cinder/autogenerated_client.go
axw, line 59: Yes, that would be awkward. Let's leave this for now.

dimitern:
@kat-co let's try merging this with the bot! :)

nova/nova.go
```diff
@@ -706,3 +707,32 @@ func (c *Client) ListAvailabilityZones() ([]AvailabilityZone, error) {
...
+
+	var resp struct {
+		VolumeAttachment *VolumeAttachment
+	}
+	requestData := goosehttp.RequestData{RespValue: &resp}
+	url := fmt.Sprintf("%s/%s/%s", apiServers, serverId, apiVolumeAttachments)
+	err := c.client.SendRequest(client.POST, "compute", url, &requestData)
+	if errors.IsNotFound(err) {
+		return nil, errors.NewNotImplementedf(
+			err, nil, "the server does not support availability zones",
```
wallyworld: I think this is a cut and paste snafu.

nova/nova.go
```diff
@@ -706,3 +707,32 @@ func (c *Client) ListAvailabilityZones() ([]AvailabilityZone, error) {
...
+}
+
+type VolumeAttachment struct {
+	Device   string `json:"device"`
+	Id       string `json:"id"`
+	ServerId string `json:"serverId"`
+	VolumeId string `json:"volumeId"`
+}
+
+func (c *Client) AttachVolume(serverId, volumeId string) (*VolumeAttachment, error) {
```
wallyworld: Can we add a comment

nova/nova.go
```diff
@@ -706,3 +707,32 @@ func (c *Client) ListAvailabilityZones() ([]AvailabilityZone, error) {
 	}
 	return resp.AvailabilityZoneInfo, nil
 }
+
+func (c *Client) ListExtensions() {
```
wallyworld: Is this intentional?

nova/nova.go
kat-co: Correct you are!

nova/nova.go
kat-co: Added comments on both VolumeAttachment and AttachVolume.

nova/nova.go
kat-co: Nope; removed.

kat-co:
$$merge$$

PR #7, opened 2015-04-09
nova: fix volume-attachment req/resp serialisation

axw:
Fix the serialisation of the new volume attachment
methods in the Nova API. There were some missing
wrappers on the types passed in/out of the API.

Tested live with Juju.


wallyworld:
$$merge$$

axw:
Thanks.

PR #8, opened 2015-04-28
Enable patching out of test service server listing

bz2:
Adds a function patch point in matchServers which can be used
to return an error when listing all servers using the http
api.

Also makes the case where a bad filter is passed in behave
more like nova does by returning an error rather than an
empty set of servers.


dimitern:
LGTM

bz2:
$$merge$$

PR #9, opened 2015-05-22
nova: add Metadata to RunServerOpts

axw:
This commit adds the Metadata field to
RunServerOpts, and updates the test server
to extract the parameter and add it to
server details.


bz2:
Looks good - but I think metadata should actually be `map[string]string` not `map[string]interface{}` though I'm struggling to find definitive docs from the openstack side.

The v2 compute api docs are explict about the key/value being strings, and has an extra /metadata endpoint for extra manipulation after server creation:

http://developer.openstack.org/api-ref-compute-v2.html#compute_server_metadata

That seems to be gone in v2.1 though adding at server creation time is still fine.

I tried giving a non-string key to the /metadata endpoint on canonistack, and got it to blow up:

```
{"computeFault": {"message": "The server has either erred or is incapable of performing the requested operation.", "code": 500}}
```

axw:
@bz2 Thanks, I've changed it to `map[string]string`.

wallyworld:
LGTM

PR #12, opened 2015-07-22
cinder: correct type for Volume.Metadata

axw:
Fix the type of Volume.Metadata. The type was wrongly deduced by wadl2go.

Also, a couple of drive-by fixes found when running live tests.


wallyworld:
LGTM

axw:
$$merge$$

PR #13, opened 2015-07-22
nova, cinder: volume attachments fixes

axw:
Two fixes:
- cinder.Volume.Attachments should have a slice-of-struct
  type, not []interface{}. This is done now. The JSON
  schema has been verified with live data.
- nova.Client.DetachVolume should expect 202 from the
  server upon successful detach. This is according to
  the docs, and to reality (found a bug testing Juju).


wallyworld:
LGTM

axw:
$$merge$$

PR #15, opened 2015-08-27
Initial support for v3 userpass authentication.

mhilton:
This is support for v3 authentication, currently only using the default domain.

### DIFF
--- a/identity/identity.go
+++ b/identity/identity.go
@@ -15,9 +15,10 @@ import (
 type AuthMode int
 
 const (
-	AuthLegacy   = AuthMode(iota) // Legacy authentication
-	AuthUserPass                  // Username + password authentication
-	AuthKeyPair                   // Access/secret key pair authentication
+	AuthLegacy     = AuthMode(iota) // Legacy authentication
+	AuthUserPass                    // Username + password authentication
+	AuthKeyPair                     // Access/secret key pair authentication
+	AuthV3UserPass                  //Username + password authentication (v3 API)
 )
 
 func (a AuthMode) String() string {
@@ -28,6 +29,8 @@ func (a AuthMode) String() string {
 		return "Legacy Authentication"
 	case AuthUserPass:
 		return "Username/password Authentication"
+	case AuthV3UserPass:
+		return "Username/password (Version 3) Authentication"
 	}
 	panic(fmt.Errorf("Unknown athentication type: %d", a))
 }
@@ -116,5 +119,7 @@ func NewAuthenticator(authMode AuthMode, httpClient *goosehttp.Client) Authentic
 		return &UserPass{client: httpClient}
 	case AuthKeyPair:
 		return &KeyPair{client: httpClient}
+	case AuthV3UserPass:
+		return &V3UserPass{client: httpClient}
 	}
 }
--- a/identity/v3userpass.go
+++ b/identity/v3userpass.go
@@ -1,0 +1,182 @@
+package identity
+
+import (
+	"fmt"
+	"net/http"
+	"time"
+
+	goosehttp "gopkg.in/goose.v1/http"
+)
+
+// v3AuthWrapper wraps the v3AuthRequest to perform v3 authentication.
+type v3AuthWrapper struct {
+	Auth v3AuthRequest `json:"auth"`
+}
+
+// v3AuthRequest contains the authentication request.
+type v3AuthRequest struct {
+	Identity v3AuthIdentity `json:"identity"`
+	Scope    *v3AuthScope   `json:"scope,omitempty"`
+}
+
+// v3AuthIdentity contains the identity portion of an authentication
+// request.
+type v3AuthIdentity struct {
+	Methods  []string        `json:"methods"`
+	Password *v3AuthPassword `json:"password,omitempty"`
+	Token    *v3AuthToken    `json:"token,omitempty"`
+}
+
+// v3AuthPassword contains a password authentication request.
+type v3AuthPassword struct {
+	User v3AuthUser `json:"user"`
+}
+
+// v3AuthUser contains the user part of a password authentication
+// request..
+type v3AuthUser struct {
+	Domain   *v3AuthDomain `json:"domain,omitempty"`
+	ID       string        `json:"id,omitempty"`
+	Name     string        `json:"name,omitempty"`
+	Password string        `json:"password"`
+}
+
+// v3AuthDomain contains a domain definition of an authentication
+// request.
+type v3AuthDomain struct {
+	ID   string `json:"id,omitempty"`
+	Name string `json:"name,omitempty"`
+}
+
+// v3AuthToken contains the token to use for token authentication.
+type v3AuthToken struct {
+	ID string `json:"id"`
+}
+
+// v3AuthScope contains the scope of the authentication request.
+type v3AuthScope struct {
+	Domain  *v3AuthDomain  `json:"domain,omitempty"`
+	Project *v3AuthProject `json:"project,omitempty"`
+}
+
+// v3AuthProject contains the project scope for the authentication
+// request.
+type v3AuthProject struct {
+	Domain *v3AuthDomain `json:"domain,omitempty"`
+	ID     string        `json:"id,omitempty"`
+	Name   string        `json:"name,omitempty"`
+}
+
+/// / V3UserPass is an Authenticator that will perform username + password
+// authentication using the v3 protocol.
+type V3UserPass struct {
+	client *goosehttp.Client
+}
+
+// Auth performs a v3 username + password authentication request using
+// the values supplied in creds.
+func (u *V3UserPass) Auth(creds *Credentials) (*AuthDetails, error) {
+	if u.client == nil {
+		u.client = goosehttp.New()
+	}
+	auth := v3AuthWrapper{
+		Auth: v3AuthRequest{
+			Identity: v3AuthIdentity{
+				Methods: []string{"password"},
+				Password: &v3AuthPassword{
+					User: v3AuthUser{
+						Domain: &v3AuthDomain{
+							ID: "default",
+						},
+						Name:     creds.User,
+						Password: creds.Secrets,
+					},
+				},
+			},
+		},
+	}
+	if creds.TenantName != "" {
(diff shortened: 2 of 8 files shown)


identity/v3userpass.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+// v3AuthDomain contains a domain definition of an authentication
+// request.
+type v3AuthDomain struct {
+	ID   string `json:"id,omitempty"`
+	Name string `json:"name,omitempty"`
+}
+
+// v3AuthToken contains the token to use for token authentication.
```
urosj, line 51: I'm not familiar with the details, but does this cover all types of tokens? the uuid, pki and fernet one?

urosj, line 51: I think we're fine.

urosj:
code LGTM, I'm not entirely convinced the AuthV3UserPass is the right way and that maybe we should introduce keystone api version attribute instead.

identity/identity.go
```diff
@@ -15,9 +15,10 @@ import (
 type AuthMode int
 
 const (
-	AuthLegacy   = AuthMode(iota) // Legacy authentication
-	AuthUserPass                  // Username + password authentication
-	AuthKeyPair                   // Access/secret key pair authentication
+	AuthLegacy     = AuthMode(iota) // Legacy authentication
+	AuthUserPass                    // Username + password authentication
+	AuthKeyPair                     // Access/secret key pair authentication
+	AuthV3UserPass                  //Username + password authentication (v3 API)
```
dimitern, line 21: Add a space between // and Username please.

identity/identity.go
```diff
@@ -28,6 +29,8 @@ func (a AuthMode) String() string {
 		return "Legacy Authentication"
 	case AuthUserPass:
 		return "Username/password Authentication"
+	case AuthV3UserPass:
+		return "Username/password (Version 3) Authentication"
```
dimitern, line 33: Username/password Authentication (Version 3) ?

To minimize confusion, we should perhaps clarify which version the original AuthUserPass is using.

identity/v3userpass.go
```diff
@@ -0,0 +1,182 @@
...
+
+// v3AuthProject contains the project scope for the authentication
+// request.
+type v3AuthProject struct {
+	Domain *v3AuthDomain `json:"domain,omitempty"`
+	ID     string        `json:"id,omitempty"`
+	Name   string        `json:"name,omitempty"`
+}
+
+/// / V3UserPass is an Authenticator that will perform username + password
```
dimitern, line 70: Remove the extra "/ /" before V3UserPass.

identity/v3userpass.go
```diff
@@ -0,0 +1,182 @@
...
+
+type v3TokenCatalog struct {
+	ID        string            `json:"id"`
+	Type      string            `json:"type"`
+	Endpoints []v3TokenEndpoint `json:"endpoints"`
+}
+
+type v3TokenEndpoint struct {
+	ID        string `json:"id"`
+	Region    string `json:"region"`
```
dimitern, line 132: According to the official API docs - http://developer.openstack.org/api-ref-identity-v3.html
this should be "region_id" and RegionID respectively.

identity/v3userpass.go
```diff
@@ -0,0 +1,182 @@
...
+	URL       string `json:"url"`
+	Interface string `json:"interface"`
+}
+
+type v3TokenProject struct {
+	ID string `json:"id"`
```
dimitern, line 138: Why omit the optional domain section here (see the same link above - esp. the examples)?

identity/v3userpass.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+type v3TokenUser struct {
+	ID string `json:"id"`
```
dimitern, line 142: Ditto?

identity/v3userpass.go
```diff
@@ -0,0 +1,182 @@
...
+	Token v3Token `json:"token"`
+}
+
+type v3Token struct {
+	Expires time.Time        `json:"expires_at"`
+	Issued  time.Time        `json:"issued_at"`
+	Methods []string         `json:"methods"`
+	Catalog []v3TokenCatalog `json:"catalog"`
+	Project v3TokenProject   `json:"project"`
+	User    v3TokenUser      `json:"user"`
```
dimitern, line 121: How about Roles?

identity/v3userpass.go
```diff
@@ -0,0 +1,182 @@
...
+		},
+	}
+	if err := c.JsonRequest("POST", url, "", &req, nil); err != nil {
+		return nil, err
+	}
+	tok := req.RespHeaders.Get("X-Subject-Token")
+	if tok == "" {
+		return nil, fmt.Errorf("authentication failed")
+	}
+	rsu := make(map[string]ServiceURLs, len(resp.Token.Catalog))
```
dimitern, line 162: If I'm reading the API docs correctly, this code only caters for the case when "catalog" is returned inside a "token", but not when there is only "token", without "catalog" inside it.

identity/v3userpass_test.go
```diff
@@ -0,0 +1,88 @@
...
+		Secrets:    "secrets",
+		TenantName: "tenant",
+	}
+	auth, err := l.Auth(&creds)
+	c.Assert(err, gc.IsNil)
+	c.Assert(auth.Token, gc.Equals, userInfo.Token)
+	c.Assert(auth.TenantId, gc.Equals, userInfo.TenantId)
+}
+
+func (s *V3UserPassTestSuite) TestAuthWithCatalog(c *gc.C) {
```
dimitern, line 48: Please, add another test where catalog is missing, but token is present (see earlier comment).

testservices/identityservice/userpass_test.go
```diff
@@ -56,7 +56,7 @@ var authTemplate = `{
 }`
 
 func userPassAuthRequest(URL, user, key string) (*http.Response, error) {
-	client := &http.DefaultClient
```
dimitern, line 59: Why? It's quite intentional to use a pointer here, instead of a copy.
If you needed to copy http.DefaultClient in order for the tests to pass, something is _wrong_.

testservices/identityservice/legacy_test.go
```diff
@@ -30,7 +30,7 @@ func (s *LegacySuite) setupLegacy(user, secret string) (token, managementURL str
 }
 
 func LegacyAuthRequest(URL, user, key string) (*http.Response, error) {
-	client := &http.DefaultClient
```
dimitern, line 33: Why (see below)?

testservices/identityservice/v3userpass.go
```diff
@@ -0,0 +1,198 @@
...
+
+// Implement the v3 User Pass form of identity (Keystone)
+
+type V3UserPassRequest struct {
+	Auth struct {
+		Identity struct {
+			Methods  []string `json:"methods"`
+			Password struct {
+				User struct {
+					Name     string `json:"name"`
```
dimitern, line 21: How about ID?

testservices/identityservice/v3userpass.go
```diff
@@ -0,0 +1,198 @@
...
+					Password string `json:"password"`
+				} `json:"user"`
+			} `json:"password"`
+		} `json:"identity"`
+		Scope struct {
+			Project struct {
+				Name string `json:"name"`
```
dimitern, line 28: Ditto.

testservices/identityservice/v3userpass.go
```diff
@@ -0,0 +1,198 @@
...
+			} `json:"project"`
+		} `json:"scope"`
+	} `json:"auth"`
+}
+
+type V3Endpoint struct {
+	Interface string `json:"interface"`
+	Region    string `json:"region"`
```
dimitern, line 36: RegionID / region_id (see first few comments)

testservices/identityservice/v3userpass.go
```diff
@@ -0,0 +1,198 @@
...
+	Expires time.Time   `json:"expires_at"`
+	Issued  time.Time   `json:"issued_at"`
+	Methods []string    `json:"methods"`
+	Catalog []V3Service `json:"catalog,omitempty"`
+	Project *v3Project  `json:"project,omitempty"`
+	User    struct {
+		ID   string `json:"id"`
+		Name string `json:"name"`
+	} `json:"user"`
+}
```
dimitern, line 56: Roles?

testservices/identityservice/v3userpass.go
```diff
@@ -0,0 +1,198 @@
+package identityservice
```
dimitern, line 1: It looks like a lot of duplicated code from userpass.go - why not embed UserPass within V3UserPass and reuse it, overriding only where necessary?

testservices/identityservice/v3userpass.go
```diff
@@ -0,0 +1,198 @@
...
+	res.Methods = []string{"password"}
+	res.Catalog = u.services
+	res.User.ID = userInfo.Id
+
+	return &res, nil
+}
+
+// setupHTTP attaches all the needed handlers to provide the HTTP API.
+func (u *V3UserPass) SetupHTTP(mux *http.ServeMux) {
+	mux.Handle("/v3/auth/tokens", u)
```
dimitern, line 197: NOTE: Because V3UserPass handles a more specific URL paths, it needs to be registered before the older UserPass.

testservices/identityservice/v3userpass_test.go
```diff
@@ -0,0 +1,146 @@
+package identityservice
```
dimitern, line 1: Similarly here - to reduce the code duplication I'd suggest extracting a BaseUserPassSuite with the common code, then embed it into UserPassSuite and V3UserPassSuite.

dimitern:
NOT LGTM I'm afraid - it needs some clean up first and reduction of the amount of duplicated code.

testservices/identityservice/userpass_test.go
mhilton, line 59: It was indeed to make the tests compile.
http.DefaultClient is already a pointer (http://golang.org/pkg/net/http/#pkg-variables)

wallyworld:
Closing this PR as support for v3 auth has landed recently.

PR #16, opened 2015-11-04
Use endpoint from keystone in cinder client

bz2:
Fixes LP:1512399, "Get https:... local error: record overflow".

Previously the autogenerated code was inventing URLs to match a
particular form, then updating just the host to match the real
endpoint. This is insufficient if either the scheme is not https
or if there are other leading path components.

The autogenerated client code has already been manually edited,
this change drops the pretense we can regenerate it, the master
copy of the WADL files no longer works.

There are some minor api changes here, the signature of NewClient
has been changed, but the Basic entry point used by the openstack
provider in juju is untouched. Also, the use of v2 paths is no
longer enforced by the code through simply ignoring the keystone
return values. Callers must verify the endpoint they are using
is v2 if they require v2 functionality.


cinder/autogenerated_client.go
```diff
@@ -1,14 +1,11 @@
 // Copyright 2015 Canonical Ltd.
 // Licensed under the LGPLv3, see LICENCE file for details.
 
-// This file is not intended to be updated by hand, rather it should
-// be regenerated every time this library is built. The WADLs were
-// originally sourced from:
+// This file was initially autogenerated from the upstream wadl file, but has
```
axw, line 4: This makes me sad, but the WADL format really isn't complete enough to auto-generate everything anyway. Pity.

axw:
LGTM

bz2:
$$merge$$

PR #17, opened 2015-12-21
add update for secgroup and server

perrito666:
- Support updating of name and description for security groups as
  described in the Openstack API
  http://developer.openstack.org/api-ref-networking-v2-ext.html#updateSecGrou
- Added the capability to update server name as per the API available
  for openstack compute (there are two different update possibilities,
  the only one implemented is the one that supports name only)
  http://developer.openstack.org/api-ref-compute-v2.html#updateServer


axw:
LGTM

perrito666:
$$merge$$

bz2:
$$merge$$

PR #25, opened 2016-10-10
Update goose to find endpoints by requesting version information befo…

hmlanigan:
…re calls

This is a step on the way to API updates to include neutron replacements for deprecated nova API calls.  There is a corresponding commit required by juju.

The POC was from: https://github.com/axw/goose/commit/a7c4a4ef9d936eb113e6ab1c0d2af4612467cedb

Rackspace and OpenStack get the same results with swift testservice (local and live).  Nova testservice has a bug so the live tests won't run there.  Identity (live) testservice tests have a permission problem I haven't been able to work out yet, that doesn't occur in real life.  juju deploy with this code and the corresponding juju PR works for both Rackspace and OpenStack.

### DIFF
--- a/client/apiversion.go
+++ b/client/apiversion.go
@@ -1,0 +1,213 @@
+package client
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+	"path"
+	"strconv"
+	"strings"
+
+	goosehttp "gopkg.in/goose.v1/http"
+)
+
+type apiVersion struct {
+	major int
+	minor int
+}
+
+type apiVersionInfo struct {
+	Version apiVersion       `json:"id"`
+	Links   []apiVersionLink `json:"links"`
+	Status  string           `json:"status"`
+}
+
+type apiVersionLink struct {
+	Href string `json:"href"`
+	Rel  string `json:"rel"`
+}
+
+type apiURLVersion struct {
+	rootURL          url.URL
+	serviceURLSuffix string
+	versions         []apiVersionInfo
+}
+
+// getAPIVersionURL returns a full formed serviceURL based on the API version requested,
+// the rootURL and the serviceURLSuffix.  If there is no match to the requested API
+// version an error is returned.  If only the major number is defined for the requested
+// version, the first match found is returned.
+func getAPIVersionURL(apiURLVersionInfo *apiURLVersion, requested apiVersion) (string, error) {
+	var match string
+	for _, v := range apiURLVersionInfo.versions {
+		if v.Version.major != requested.major {
+			continue
+		}
+		if requested.minor != -1 && v.Version.minor != requested.minor {
+			continue
+		}
+		for _, link := range v.Links {
+			if link.Rel != "self" {
+				continue
+			}
+			hrefURL, err := url.Parse(link.Href)
+			if err != nil {
+				return "", err
+			}
+			match = hrefURL.Path
+		}
+		if requested.minor != -1 {
+			break
+		}
+	}
+	if match == "" {
+		return "", fmt.Errorf("could not find matching URL")
+	}
+	versionURL := apiURLVersionInfo.rootURL
+	versionURL.Path = path.Join(versionURL.Path, match, apiURLVersionInfo.serviceURLSuffix)
+	return versionURL.String(), nil
+}
+
+func (v *apiVersion) UnmarshalJSON(b []byte) error {
+	var s string
+	if err := json.Unmarshal(b, &s); err != nil {
+		return err
+	}
+	parsed, err := parseVersion(s)
+	if err != nil {
+		return err
+	}
+	*v = parsed
+	return nil
+}
+
+// parseVersion takes a version string into the major and minor ints for an apiVersion
+// structure. The string part of the data is returned by a request to List API versions
+// send to an OpenStack service.  It is in the format "v<major>.<minor>". If apiVersion
+// is empty, return {-1, -1}, to differentiate with "v0".
+func parseVersion(s string) (apiVersion, error) {
+	if s == "" {
+		return apiVersion{-1, -1}, nil
+	}
+	s = strings.TrimPrefix(s, "v")
+	parts := strings.SplitN(s, ".", 2)
+	if len(parts) == 0 || len(parts) > 2 {
+		return apiVersion{}, fmt.Errorf("invalid API version %q", s)
+	}
+	var minor int = -1
+	major, err := strconv.Atoi(parts[0])
+	if err != nil {
+		return apiVersion{}, err
+	}
+	if len(parts) == 2 {
+		var err error
+		minor, err = strconv.Atoi(parts[1])
+		if err != nil {
+			return apiVersion{}, err
+		}
+	}
+	return apiVersion{major, minor}, nil
+}
+
+func unmarshallVersion(Versions json.RawMessage) ([]apiVersionInfo, error) {
+	// Some services respond with {"versions":[...]}, and
+	// some respond with {"versions":{"values":[...]}}.
+	var object interface{}
+	var versions []apiVersionInfo
+	if err := json.Unmarshal(Versions, &object); err != nil {
+		return versions, err
+	}
+	if _, ok := object.(map[string]interface{}); ok {
+		var valuesObject struct {
+			Values []apiVersionInfo `json:"values"`
+		}
+		if err := json.Unmarshal(Versions, &valuesObject); err != nil {
+			return versions, err
+		}
+		versions = valuesObject.Values
+	} else {
+		if err := json.Unmarshal(Versions, &versions); err != nil {
+			return versions, err
+		}
+	}
+	return versions, nil
+}
+
+// getAPIVersions returns data on the API versions supported by the specified service endpoint.
+func (c *authenticatingClient) getAPIVersions(serviceCatalogURL string) (*apiURLVersion, error) {
+	url, err := url.Parse(serviceCatalogURL)
+	if err != nil {
+		return nil, err
+	}
+
+	// Identify the version in the URL, if there is one, and record
+	// everything proceeding it. We will need to append this to the
+	// API version-specific base URL.
+	var pathParts, origPathParts []string
+	if url.Path != "/" {
+		// The rackspace object-store endpoint triggers the version removal here,
+		// so keep the original parts for object-store and container endpoints.
+		// e.g. https://storage101.dfw1.clouddrive.com/v1/MossoCloudFS_1019383
+		origPathParts = strings.Split(strings.Trim(url.Path, "/"), "/")
+		pathParts = origPathParts
+		if _, err := parseVersion(pathParts[0]); err == nil {
+			pathParts = pathParts[1:]
+		}
+		url.Path = "/"
+	}
+
+	baseURL := url.String()
+
+	// If this is an object-store serviceType, or an object-store container endpoint,
+	// there is no list version API call to make. Return a apiURLVersion which will
+	// satisfy a requested api version of "", "v1" or "v1.0"
+	if c.serviceURLs["object-store"] != "" && strings.Contains(serviceCatalogURL, c.serviceURLs["object-store"]) {
+		objectStoreLink := apiVersionLink{Href: baseURL, Rel: "self"}
+		objectStoreApiVersionInfo := []apiVersionInfo{
+			{
+				Version: apiVersion{major: 1, minor: 0},
+				Links:   []apiVersionLink{objectStoreLink},
+				Status:  "stable",
+			},
+			{
+				Version: apiVersion{major: -1, minor: -1},
+				Links:   []apiVersionLink{objectStoreLink},
+				Status:  "stable",
+			},
+		}
+		return &apiURLVersion{*url, strings.Join(origPathParts, "/"), objectStoreApiVersionInfo}, nil
+	}
+
+	// make sure we haven't already received the version info
+	c.apiVersionMu.Lock()
+	defer c.apiVersionMu.Unlock()
+	if apiInfo, ok := c.apiURLVersions[baseURL]; ok {
+		return apiInfo, nil
+	}
+
+	var raw struct {
+		Versions json.RawMessage `json:"versions"`
+	}
+	requestData := &goosehttp.RequestData{
+		RespValue: &raw,
+		ExpectedStatus: []int{
+			http.StatusOK,
+			http.StatusMultipleChoices,
+		},
+	}
+	if err := c.sendRequest("GET", baseURL, c.Token(), requestData); err != nil {
+		return nil, err
+	}
+
+	versions, err := unmarshallVersion(raw.Versions)
+	if err != nil {
+		return nil, err
+	}
+
+	// save this info, so we don't have to get it again
+	apiURLVersionInfo := &apiURLVersion{*url, strings.Join(pathParts, "/"), versions}
+	c.apiURLVersions[baseURL] = apiURLVersionInfo
+
+	return apiURLVersionInfo, nil
+}
--- a/client/apiversion_test.go
+++ b/client/apiversion_test.go
@@ -1,0 +1,191 @@
+package client_test
+
+import (
+	"fmt"
+	"net/http"
+	"strconv"
+
+	gc "gopkg.in/check.v1"
+)
+
+type versionHandler struct {
+	authBody string
+	port     string
+}
+
+type makeServiceURLTest struct {
+	serviceType string
+	version     string
+	parts       []string
+	success     bool
+	URL         string
+	err         string
+}
+
+func (s *localLiveSuite) makeServiceURLTests() []makeServiceURLTest {
+	return []makeServiceURLTest{
+		{
+			serviceType: "compute",
+			version:     "v2.1",
+			parts:       []string{"foo", "bar/"},
+			success:     true,
+			URL:         "http://localhost:%s/v2.1/foo/bar/",
+		},
+		{
+			serviceType: "compute",
+			version:     "v2.0",
+			parts:       []string{},
+			success:     true,
+			URL:         "http://localhost:%s/v2.0",
+		},
+		{
+			serviceType: "compute",
+			version:     "v2.0",
+			parts:       []string{"foo", "bar/"},
+			success:     true,
+			URL:         "http://localhost:%s/v2.0/foo/bar/",
+		},
+		{
+			serviceType: "compute",
+			version:     "v2",
+			parts:       []string{"foo", "bar/"},
+			success:     true,
+			URL:         "http://localhost:%s/v2.1/foo/bar/",
+		},
+		{
+			serviceType: "object-store",
+			version:     "",
+			parts:       []string{"foo", "bar"},
+			success:     true,
+			URL:         "http://localhost:%s/swift/v1/foo/bar",
+		},
+		{
+			serviceType: "object-store",
+			version:     "q2.0",
+			parts:       []string{"foo", "bar/"},
+			success:     false,
+			err:         "strconv.ParseInt: parsing \"q2\": invalid syntax",
+		},
+		{
+			serviceType: "object-store",
+			version:     "v1.7",
+			parts:       []string{"foo", "bar/"},
+			success:     false,
+			err:         "could not find matching URL",
+		},
+		{
+			serviceType: "juju-container-test",
+			version:     "v1",
+			parts:       []string{"foo", "bar/"},
+			success:     true,
+			URL:         "http://localhost:%s/swift/v1/foo/bar/",
+		},
+		{
+			serviceType: "juju-container-test",
+			version:     "v0",
+			parts:       []string{"foo", "bar/"},
+			success:     false,
+			err:         "could not find matching URL",
+		},
+		{
+			serviceType: "juju-container-test",
+			version:     "",
+			parts:       []string{"foo", "bar/"},
+			success:     true,
+			URL:         "http://localhost:%s/swift/v1/foo/bar/",
+		},
+		{
+			serviceType: "compute",
+			version:     "",
+			parts:       []string{},
+			success:     false,
+			err:         "could not find matching URL",
+		},
+		{
+			serviceType: "compute",
+			version:     "v25.4",
+			parts:       []string{},
+			success:     false,
+			err:         "could not find matching URL",
+		},
+		{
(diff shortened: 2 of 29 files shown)


client/client.go
```diff
@@ -163,44 +167,213 @@ func makeURL(base string, parts []string) string {
...
-	url, err := c.MakeServiceURL(svcType, []string{apiCall})
+	url, err := c.MakeServiceURL(svcType, apiVersion, []string{apiCall})
 	if err != nil {
 		return
 	}
 	return c.sendRequest(method, url, c.Token(), requestData)
 }
 
-func (c *authenticatingClient) MakeServiceURL(serviceType string, parts []string) (string, error) {
+func (c *authenticatingClient) MakeServiceURL(serviceType, apiVersion string, parts []string) (string, error) {
```
axw, line 208: can you please add caching of the version-specific URLs? as it is, this will query the versions every single API request

axw (changes requested):

client/client.go
hmlanigan: added to client.go: getAPIVersions()

client/client.go
```diff
@@ -163,44 +170,231 @@ func makeURL(base string, parts []string) string {
...
 	if !ok {
 		return "", errors.New("no endpoints known for service type: " + serviceType)
 	}
-	return makeURL(url, parts), nil
+	if serviceType == "object-store" {
+		// object-store has no versioning capability.
+		return makeURL(serviceURL, parts), nil
+	}
+	if apiVersion == "" {
+		// some tests don't include a version
```
wallyworld: Let's not add special code to account for poorly constructed tests. If the apiVersion is required, any tests should be updated accordingly.

client/client.go
```diff
@@ -163,44 +170,231 @@ func makeURL(base string, parts []string) string {
...
+		return makeURL(serviceURL, parts), nil
+	}
+	if strings.Contains(serviceURL, c.serviceURLs["object-store"]) {
```
wallyworld: Isn't this handling of object-store service type done up above?

client/client.go
```diff
@@ -163,44 +170,231 @@ func makeURL(base string, parts []string) string {
...
+		// a container endpoint won't have a version
+		return makeURL(serviceURL, parts), nil
+	}
+	requestedVersion, err := parseVersion(apiVersion)
```
wallyworld, line 216: can we move this down next to getVersionURL where it is used

client/client.go
```diff
@@ -163,44 +170,231 @@ func makeURL(base string, parts []string) string {
...
+		return "", err
+	}
+	serviceURL, err = getAPIVersionURL(rootURL, serviceURLSuffix, versions, requestedVersion)
+	if err != nil {
+		return "", err
+	}
+	return makeURL(serviceURL, parts), nil
+}
+
+func getAPIVersionURL(rootURL *url.URL, serviceURLSuffix string, versions []apiVersionInfo, requested apiVersion) (string, error) {
```
wallyworld: i think there's enough functionality and parameters here to warrant a comment

client/client.go
```diff
@@ -163,44 +170,231 @@ func makeURL(base string, parts []string) string {
...
+	if match == "" {
+		return "", fmt.Errorf("could not find matching URL")
+	}
+	versionURL := *rootURL
+	versionURL.Path = path.Join(versionURL.Path, match, serviceURLSuffix)
+	return versionURL.String(), nil
+}
+
+type apiVersion struct {
+	major int
```
wallyworld: if we are going to unmarshal from json, our convention is to ensure there are json tags defined

client/client.go
```diff
@@ -163,44 +170,231 @@ func makeURL(base string, parts []string) string {
...
+	minor int
+}
+
+func (v *apiVersion) UnmarshalJSON(b []byte) error {
```
wallyworld: All this api version stuff probably is best placed in a separate file like apiversion.go

client/client.go
```diff
@@ -163,44 +170,231 @@ func makeURL(base string, parts []string) string {
...
+		if _, err := parseVersion(pathParts[0]); err == nil {
+			pathParts = pathParts[1:]
+		}
+		url.Path = "/"
+	}
+
+	baseURL := url.String()
+
+	// make sure we haven't already received the version info
+	apiURLVersion, ok := c.apiURLVersions[baseURL]
```
wallyworld: This isn't go routine safe and should be protected by a mutex

client/client.go
```diff
@@ -163,44 +170,231 @@ func makeURL(base string, parts []string) string {
...
+	if ok {
```
wallyworld: this can be inlined with the apiURLVersions look up

client/client.go
```diff
@@ -163,44 +170,231 @@ func makeURL(base string, parts []string) string {
...
+		return apiURLVersion.rootURL, apiURLVersion.serviceURLSuffix, apiURLVersion.versions, nil
```
wallyworld: instead of returning (rootURL, serviceURLSuffix, versions) why not just return the apiURLVersion struct itself

client/client.go
```diff
@@ -163,44 +170,231 @@ func makeURL(base string, parts []string) string {
...
+	}
+	requestData := &goosehttp.RequestData{
+		RespValue: &raw,
+		ExpectedStatus: []int{
+			http.StatusOK,
+			http.StatusMultipleChoices,
+		},
+	}
+	if err := c.sendRequest("GET", baseURL, c.Token(), requestData); err != nil {
+		return nil, "", nil, err
```
wallyworld: yeah, we should just return *apiURLVersion
then this would be return nil, err

client/client.go
```diff
@@ -163,44 +170,231 @@ func makeURL(base string, parts []string) string {
...
+	}
+
+	// Some services respond with {"versions":[...]}, and
+	// some respond with {"versions":{"values":[...]}}.
+	var object interface{}
+	var versions []apiVersionInfo
+	if err := json.Unmarshal(raw.Versions, &object); err != nil {
```
wallyworld: all this ummarshal stuff should be in a separate helper function, possible in that separate file mentioned elsewhere

testservices/cmd/main.go
```diff
@@ -67,5 +67,6 @@ func main() {
 	for _, u := range users.users {
 		p.AddUser(u.user, u.secret, "tenant")
 	}
+	fmt.Printf("main(): %s\n", serveAddr)
```
wallyworld: delete me? or log at debug level?

testservices/novaservice/service.go
```diff
@@ -143,6 +150,53 @@ func New(hostURL, versionPath, tenantId, region string, identityService, fallbac
 	return novaService
 }
 
+func NewNovaOnly(cred *identity.Credentials, auth identity.AuthMode) (*Nova, []string) {
```
wallyworld: nitpic - all exported methods should have a comment, here and a few other places

testservices/novaservice/service_http.go
```diff
@@ -167,8 +167,9 @@ The resource could not be found.
 	}
 	errNoVersion = &errorResponse{
 		http.StatusOK,
-		`{"versions": [{"status": "CURRENT", "updated": "2011-01-21` +
-			`T11:33:21Z", "id": "v2.0", "links": [{"href": "$ENDPOINT$", "rel": "self"}]}]}`,
+		`{"versions": [` +
+			`{"id": "v2.0", "links": [{"href": "v2", "rel": "self"}], "status": "SUPPORTED", "updated": "2011-01-21T11:33:21Z"}]}`,
+		//`{"id": "v2.1", "links": [{"href": "v2.1", "rel": "self"}], "status": "CURRENT", "updated": "2013-07-23T11:33:21Z"}]}`,
```
wallyworld: delete me?

wallyworld (changes requested):
There needs to be unit test coverage of the getAPIVersions and getAPIVersionURL methods. There's a lot of logic / corner cases there, eg
// Some services respond with {"versions":[...]},
// some respond with {"versions":{"values":[...]}}.

Whether those methods are tested explicitly, or via curated calls to MakeServiceURL doesn't really matter, although the latter may be preferred.

client/client.go
```diff
@@ -163,44 +170,231 @@ func makeURL(base string, parts []string) string {
 	return base + strings.Join(parts, "/")
 }
 
-func (c *client) MakeServiceURL(serviceType string, parts []string) (string, error) {
+func (c *client) MakeServiceURL(serviceType, apiVersion string, parts []string) (string, error) {
 	return makeURL(c.baseURL, parts), nil
```
wallyworld, line 171: dumb question - does a non authenticating client not require apiVersion?

hmlanigan:
In theory, I agree.  In practice, I’m not sure there are OpenStack deployments without username/password or token required?  None of the current goose, nor juju openstack provider test code uses a non authenticating client.  Only the goose client tests do in cases to do with containers, Swift doesn’t use versions.

I need some input about what is supported with openstack deployments used by juju.


On Oct 11, 2016, at 10:03 PM, Ian Booth <notifications@github.com<mailto:notifications@github.com>> wrote:


@wallyworld commented on this pull request.

________________________________

In client/client.go<https://github.com/go-goose/goose/pull/25#pullrequestreview-3798096>:

> @@ -163,44 +170,231 @@ func makeURL(base string, parts []string) string {
        return base + strings.Join(parts, "/")
 }

-func (c *client) MakeServiceURL(serviceType string, parts []string) (string, error) {
+func (c *client) MakeServiceURL(serviceType, apiVersion string, parts []string) (string, error) {
        return makeURL(c.baseURL, parts), nil


dumb question - does a non authenticating client not require apiVersion?

client/client.go
hmlanigan: No, users can add endpoints with whatever URL they want, with a service type of whatever they want.  One way to pass simplestreams data to Juju is to put it in a Swift container and add an endpoint. Juju explicitly looks for an endpoint with service type "product-streams", and it's tests use other made up names.  We do know that Swift containers are used, so the root url will be the same as the "object-store" url.

client/client.go
hmlanigan: done

client/client.go
hmlanigan: done

client/client.go
hmlanigan: In theory, I agree.  In practice, I’m not sure there are OpenStack deployments without username/password or token required?  None of the current goose, nor juju openstack provider test code uses a non authenticating client.  Only the goose client tests do in cases to do with containers, Swift doesn’t use versions.

I need some input about what is supported with openstack deployments used by juju.

testservices/novaservice/service.go
hmlanigan: I'll add comments to the exported functions I changed the functionality of.

client/client.go
hmlanigan: done

client/client.go
hmlanigan: done

client/client.go
hmlanigan: done

testservices/cmd/main.go
hmlanigan: removed.

client/client.go
hmlanigan: per discussion with wallyworld - invalid

client/client.go
hmlanigan: done

client/client.go
hmlanigan: done

client/client.go
hmlanigan: resolved

client/client.go
hmlanigan: removed, issue resolved with check for containers.

testservices/novaservice/service_http.go
hmlanigan: done.

client/client.go
wallyworld: yep, my bad, sorry

client/client.go
wallyworld: Juju itself only uses the authenticating client.

client/apiversion.go
```diff
@@ -0,0 +1,183 @@
...
+		versions = valuesObject.Values
+	} else {
+		if err := json.Unmarshal(Versions, &versions); err != nil {
+			return versions, err
+		}
+	}
+	return versions, nil
+}
+
+// Given a service endpoint, return data on the versions suported by the service
```
wallyworld: sorry to be picky - the comment should start with the method name and end with a period. eg
// getAPIVersions returns data on the versions supported by the service at the specified endpoint.

Same comment applies to the getAPIVersionURL method above

client/apiversion.go
```diff
@@ -0,0 +1,183 @@
...
+// and their urls
+func (c *authenticatingClient) getAPIVersions(serviceCatalogURL string) (*apiURLVersion, error) {
+	c.apiVersionMu.Lock()
```
wallyworld: This can be moved down to line 154 just before we first start using the apiURLVersions map.
That way we don't prematurely grab the lock

client/apiversion.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+	parsed, err := parseVersion(s)
+	if err != nil {
+		return err
+	}
+	*v = parsed
+	return nil
+}
+
+func parseVersion(s string) (apiVersion, error) {
```
wallyworld, line 89: To aid the reader, a comment here saying that this function parses version strings of the form "v<major>", eg "v2", or "v<major>.<minor>, eg "v2.1"" would be great. Or words to that effect.

wallyworld (commented):
Thanks for all the changes. Just add those unit tests as discussed and it should be good.

client/apiversion.go
hmlanigan: updated

client/apiversion.go
hmlanigan: done.

client/apiversion.go
hmlanigan: updated.

client/local_test.go
```diff
@@ -97,18 +98,27 @@ func (s *localLiveSuite) SetUpSuite(c *gc.C) {
...
+	if s.authMode != identity.AuthLegacy {
+		s.service.SetupHTTP(nil)
+	}
 	s.LiveTests.SetUpSuite(c)
 }
 
 func (s *localLiveSuite) TearDownSuite(c *gc.C) {
 	s.LiveTests.TearDownSuite(c)
 	s.HTTPSuite.TearDownSuite(c)
+	s.service.Stop()
```
wallyworld: The service is only started for authMode = AuthLegacy, so stopping it unconditionally doesn't make sense?
It seems that now that there's differences between legacy and non legacy auth, one option is to have a  2 suites, one for legacy auth and one for non-legacy auth. These suites would embed the localLiveSuite suite. The s.service would be moved to legacyLocalLiveSuite and TestAuthenticate would be moved off localLiveSuite and separate implementationd added to each new suite.

client/apiversion_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"net/http"
+	"strconv"
+)
+
+type versionHandler struct {
+	authBody string
+	port     string
+}
+
+type MakeServiceURLTest struct {
```
wallyworld: does this need to be exported?

client/apiversion_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+const authValuesInformationBody = `{"versions": {"values": [` +
+	`{"status": "stable", "updated": "2015-03-30T00:00:00Z", "media-types": [{"base": "application/json", "type": "application/vnd.openstack.identity-v3+json"}], "id": "v3.4", "links": [{"href": "%s/v3/", "rel": "self"}]},` +
+	`{"status": "stable", "updated": "2014-04-17T00:00:00Z", "media-types": [{"base": "application/json", "type": "application/vnd.openstack.identity-v2.0+json"}], "id": "v2.0", "links": [{"href": "%s/v2.0/", "rel": "self"}, {"href": "http://docs.openstack.org/", "type": "text/html", "rel": "describedby"}]},` +
+	`{"status": "stable", "updated": "2015-03-30T00:00:00Z", "media-types": [{"base": "application/json", "type": "application/vnd.openstack.identity-v3+json"}], "id": "v2.1", "links": [{"href": "%s/v2.1/", "rel": "self"}]}` +
+	`]}}`
+
+func (vh *versionHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	//var body []byte
```
wallyworld: delete me

client/apiversion_test.go
```diff
@@ -0,0 +1,99 @@
...
+	parts       []string
+	success     bool
+	URL         string
+	err         string
+}
+
+func (s *localLiveSuite) makeServiceURLTests() []MakeServiceURLTest {
+	return []MakeServiceURLTest{
+		{
+			serviceType: "compute",
```
wallyworld, line 28: Given object-store is special, there should be MakeServiceURL tests for URLs for things other than just compute?

client/apiversion_test.go
```diff
@@ -0,0 +1,99 @@
...
+type MakeServiceURLTest struct {
+	serviceType string
+	version     string
+	parts       []string
+	success     bool
+	URL         string
+	err         string
+}
+
+func (s *localLiveSuite) makeServiceURLTests() []MakeServiceURLTest {
```
wallyworld: Perhaps this would be best declared near the test cases which make use of it. Maybe move the MakeServiceURL tests to this file?

client/apiversion_test.go
hmlanigan: no - updated

client/apiversion_test.go
hmlanigan: moved TestMakeServiceURL and TestMakeServiceURLValues to apiversion_test.go

client/apiversion_test.go
hmlanigan: done

client/apiversion_test.go
hmlanigan: added tests for "object-store" and an "object-store" container but random service type

client/apiversion_test.go
```diff
@@ -0,0 +1,164 @@
...
+			parts:       []string{"foo", "bar"},
+			success:     true,
+			URL:         "http://localhost:%s/swift/v1/foo/bar",
+		},
+		{
+			serviceType: "object-store",
+			version:     "v2.0",
+			parts:       []string{"foo", "bar/"},
+			success:     true,
+			URL:         "http://localhost:%s/swift/v1/foo/bar/",
```
wallyworld, line 81: The code comments say that container endpoints have no versioning capability and yet there's a "v1" in the expected URL. Isn't that versioning?  It seems counter intuitive that we can ask for a v2.0 endpoint and get back a v1 URL when that would fail for compute.

hmlanigan: The endpoint says v1, however the API doesn't have a call that returns a list of versions. Here's a link to the API: http://developer.openstack.org/api-ref/object-storage/

For object-storage in MakeServiceURL - the version is ignored.  As far as I'm aware there is not version 2 of the object-store (Swift) API.  It could be that they will create a new endpoint service type if one is needed.  There is already volume and volumev2 service types for Cinder.

You do need to be aware of OpenStack services and version when making calling SendRequest and/or MakeServiceURL.

wallyworld: object-store only has one endpoint. But that endpoint does have a version, "v1".
So we can just have the MakeServiceURL caller in goose (and Juju which calls goose) pass in "" or "v1" and the correct, expected result will be returned. If the caller passes in "v2", that should still error.That allows consistent and understandable behaviour of the API, and keeps the semantics consistent without the special case.

client/client.go
```diff
@@ -163,44 +167,86 @@ func makeURL(base string, parts []string) string {
...
 	}
-	url, ok := c.serviceURLs[serviceType]
+	serviceURL, ok := c.serviceURLs[serviceType]
 	if !ok {
 		return "", errors.New("no endpoints known for service type: " + serviceType)
 	}
-	return makeURL(url, parts), nil
+	requestedVersion, verErr := parseVersion(apiVersion)
+	if serviceType == "object-store" {
+		if apiVersion == "" || (verErr == nil && requestedVersion.major == 1) {
```
wallyworld: Oh, I wasn't thinking of this approach. I thought getAPIVersions() would be modified to see if the URL was for object-store and just short circuit without calling the api and return an apiURLVersionInfo with 1.0 URL in it. That to me is the better place to put the short circuit, as it keeps the special case code next to the api in question. The code here is also a lot messier as there are checks for requestedVersion.major == 1 etc which would be avoided if the short circuit were where I am suggesting.
The fact that this works is good, let's not block on that, but my preference would be for the approach I suggest here.

wallyworld (approved):
I've asked for the short circuit logic to be moved but don't want to block on that as we need to land this. My preference would be to move then logic before landing, no need to wait for another +1

hmlanigan:
The short circuit logic was moved to getAPIVersions() as requested.  Additional tests were added accordingly.

hmlanigan:
$$merge$$

axw:
$$merge$$

hmlanigan:
Updated, for object-store or object-store container endpoints, only “” or “v1” endpoints will be accepted.

On Oct 19, 2016, at 10:17 PM, Ian Booth <notifications@github.com<mailto:notifications@github.com>> wrote:

@wallyworld commented on this pull request.

---

In client/apiversion_test.gohttps://github.com/go-goose/goose/pull/25:

> -                     success:     true,
> -                       URL:         "http://localhost:%s/v2.1/foo/bar/",
> -               },
> -               {
> -                       serviceType: "object-store",
> -                       version:     "",
> -                       parts:       []string{"foo", "bar"},
> -                       success:     true,
> -                       URL:         "http://localhost:%s/swift/v1/foo/bar",
> -               },
> -               {
> -                       serviceType: "object-store",
> -                       version:     "v2.0",
> -                       parts:       []string{"foo", "bar/"},
> -                       success:     true,
> -                       URL:         "http://localhost:%s/swift/v1/foo/bar/",

object-store only has one endpoint. But that endpoint does have a version, "v1".
So we can just have the MakeServiceURL caller in goose (and Juju which calls goose) pass in "" or "v1" and the correct, expected result will be returned. If the caller passes in "v2", that should still error.That allows consistent and understandable behaviour of the API, and keeps the semantics consistent without the special case.

PR #26, opened 2016-10-26
nova: don't pass an empty string in "device"

axw:
When attaching a volume to a server, don't pass an
empty string in the "device" field of the volume
attachment. Instead, the field should be omitted.
Some versions of OpenStack will fail the request
if an empty string is passed.

Part of the fix for https://bugs.launchpad.net/juju/+bug/1636648


wallyworld (approved):

axw:
$$merge$$

PR #27, opened 2016-10-26
client: don't do version discovery when version=""

axw:
If the user specifies no API version at all, don't
perform version discovery, and just return the URL
reported by the authentication response. This is to
support Juju's use of the "product-streams" service
entry for simplestreams data.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -204,7 +204,10 @@ func (c *authenticatingClient) sendAuthRequest(
 
 // MakeServiceURL uses an endpoint matching the apiVersion for the given service type.
 // Given a major version only, the version with the highest minor will be used.
-// object-store and container service types have no versions
+//
+// object-store and container service types have no versions. For these services, the
+// caller may pass "" for apiVersion, to use the service catalogue URL without any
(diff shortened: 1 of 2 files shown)


client/client.go
```diff
@@ -213,17 +216,19 @@ func (c *authenticatingClient) MakeServiceURL(serviceType, apiVersion string, pa
...
-		return "", err
-	}
-	apiURLVersionInfo, err := c.getAPIVersions(serviceURL)
-	if err != nil {
-		return "", err
-	}
-	serviceURL, err = getAPIVersionURL(apiURLVersionInfo, requestedVersion)
-	if err != nil {
-		return "", err
+	if apiVersion != "" {
```
wallyworld: maybe invert this to avoid the indented block
if apiVersion == "" {
  return blah
}
// do version stuff
...

wallyworld (approved):

client/client.go
axw: Done.

axw:
$$merge$$

PR #28, opened 2016-11-03
nova: omit "id" and "serverId" from attach request

axw:
When POSTing to os-volume_attachments, we should not
be passing anything for "id" or "serverId". Set these
to omitempty.


axw:
See: http://developer.openstack.org/api-ref/compute/?expanded=attach-a-volume-to-an-instance-detail

wallyworld (approved):

axw:
$$merge$$

PR #29, opened 2016-11-03
client: add method to disable version discovery

axw:
Add AuthenticatingClient.SetVersionDiscoveryEnabled
so package users can disable API version discovery,
to revert back to the old behaviour.


wallyworld (approved):

axw:
$$merge$$

PR #30, opened 2016-11-07
Add Neutron pkg and test double, update Glance

hmlanigan:
Add Neutron package and test service, functions related to Networks, Subnets, FloatingIPs, SecurityGroups, SecurityGroupRules.  Neutron and Nova test services share Network, FloatingIP and SecurityGroup data for OpenStack service.

### DIFF
--- a/glance/glance.go
+++ b/glance/glance.go
@@ -1,5 +1,6 @@
 // goose/glance - Go package to interact with OpenStack Image Service (Glance) API.
-// See http://docs.openstack.org/api/openstack-image-service/2.0/content/.
+// For V2 functions see http://developer.openstack.org/api-ref/image/v2/index.html
+// For all other functions see http://developer.openstack.org/api-ref/compute/#list-images
 
 package glance
 
@@ -105,4 +106,58 @@ func (c *Client) GetImageDetail(imageId string) (*ImageDetail, error) {
 		return nil, errors.Newf(err, "failed to get details of imageId: %s", imageId)
 	}
 	return &resp.Image, nil
+}
+
+// ImageDetailV2 describes extended information about an image for API v2.0.
+type ImageDetailV2 struct {
+	Architecture    string `json:",omitempty"`
+	Checksum        string
+	ContainerFormat string `json:"container_format"`
+	CreatedAt       string `json:"created_at"`
+	DirectUrl       string
+	DiskFormat      string `json:"disk_format"`
+	File            string
+	Id              string
+	HwVifModel      string      `json:"hw_vif_model,omitempty"`
+	HwDiskModel     string      `json:"hw_disk_bus,omitempty"`
+	Locations       interface{} `json:",omitempty"`
+	MinimumDisk     int         `json:"min_disk"`
+	MinimumRAM      int         `json:"min_ram"`
+	Name            string
+	Owner           string
+	OsType          string `json:"os_type"`
+	Protected       bool
+	Schema          string
+	Self            string
+	Size            int
+	Status          string
+	Tags            []string
+	UpdatedAt       string `json:"updated_at"`
+	VirtualSize     string `json:"virtual_size"`
+	Visibility      string
+}
+
+// ListImagesV2 lists all details for available image, uses API v2.0.
+func (c *Client) ListImagesV2() ([]ImageDetailV2, error) {
+	var resp struct {
+		Images []ImageDetailV2
+	}
+	requestData := goosehttp.RequestData{RespValue: &resp}
+	err := c.client.SendRequest(client.GET, "image", "v2", apiImages, &requestData)
+	if err != nil {
+		return nil, errors.Newf(err, "failed to get list of image details (v2)")
+	}
+	return resp.Images, nil
+}
+
+// GetImageDetailV2 lists details of the specified image, uses API v2.0
+func (c *Client) GetImageDetailV2(imageId string) (*ImageDetailV2, error) {
+	var resp ImageDetailV2
+	url := fmt.Sprintf("%s/%s", apiImages, imageId)
+	requestData := goosehttp.RequestData{RespValue: &resp}
+	err := c.client.SendRequest(client.GET, "image", "v2", url, &requestData)
+	if err != nil {
+		return nil, errors.Newf(err, "failed to get details of imageId (v2): %s", imageId)
+	}
+	return &resp, nil
 }
--- a/neutron/live_test.go
+++ b/neutron/live_test.go
@@ -1,0 +1,252 @@
+package neutron_test
+
+import (
+	gc "gopkg.in/check.v1"
+	"net"
+
+	"gopkg.in/goose.v1/client"
+	"gopkg.in/goose.v1/identity"
+	"gopkg.in/goose.v1/neutron"
+)
+
+func registerOpenStackTests(cred *identity.Credentials) {
+	gc.Suite(&LiveTests{
+		cred: cred,
+	})
+}
+
+type LiveTests struct {
+	cred     *identity.Credentials
+	client   client.AuthenticatingClient
+	neutron  *neutron.Client
+	userId   string
+	tenantId string
+}
+
+func (s *LiveTests) SetUpSuite(c *gc.C) {
+	s.client = client.NewClient(s.cred, identity.AuthUserPass, nil)
+	s.neutron = neutron.New(s.client)
+}
+
+func (s *LiveTests) TearDownSuite(c *gc.C) {
+	// noop, called by local test suite.
+}
+
+func (s *LiveTests) SetUpTest(c *gc.C) {
+	// noop, called by local test suite.
+}
+
+func (s *LiveTests) TearDownTest(c *gc.C) {
+	// noop, called by local test suite.
+}
+
+func (s *LiveTests) TestFloatingIPsV2(c *gc.C) {
+	networks, err := s.neutron.ListNetworksV2()
+	c.Assert(err, gc.IsNil)
+	var netId string
+	for _, net := range networks {
+		if net.External == true {
+			netId = net.Id
+			break
+		}
+	}
+	if netId == "" {
+		c.Errorf("no valid network to create floating IP")
+	}
+	c.Assert(netId, gc.Not(gc.Equals), "")
+	ip, err := s.neutron.AllocateFloatingIPV2(netId)
+	c.Assert(err, gc.IsNil)
+	defer s.neutron.DeleteFloatingIPV2(ip.Id)
+	c.Assert(ip, gc.Not(gc.IsNil))
+	c.Check(ip.IP, gc.Not(gc.Equals), "")
+	c.Check(ip.FixedIP, gc.Equals, "")
+	c.Check(ip.Id, gc.Not(gc.Equals), "")
+	c.Check(ip.FloatingNetworkId, gc.Not(gc.Equals), "")
+	ips, err := s.neutron.ListFloatingIPsV2()
+	c.Assert(err, gc.IsNil)
+	if len(ips) < 1 {
+		c.Errorf("no floating IPs found (expected at least 1)")
+	} else {
+		found := false
+		for _, i := range ips {
+			c.Check(i.IP, gc.Not(gc.Equals), "")
+			if i.Id == ip.Id {
+				c.Check(i.IP, gc.Equals, ip.IP)
+				c.Check(i.FloatingNetworkId, gc.Equals, ip.FloatingNetworkId)
+				found = true
+			}
+		}
+		if !found {
+			c.Errorf("expected to find added floating IP: %#v", ip)
+		}
+		fip, err := s.neutron.GetFloatingIPV2(ip.Id)
+		c.Assert(err, gc.IsNil)
+		c.Check(fip.Id, gc.Equals, ip.Id)
+		c.Check(fip.IP, gc.Equals, ip.IP)
+		c.Check(fip.FloatingNetworkId, gc.Equals, ip.FloatingNetworkId)
+	}
+	err = s.neutron.DeleteFloatingIPV2(ip.Id)
+	c.Assert(err, gc.IsNil)
+	_, err = s.neutron.GetFloatingIPV2(ip.Id)
+	c.Assert(err, gc.Not(gc.IsNil))
+}
+
+func (s *LiveTests) TestListNetworksV2(c *gc.C) {
+	networks, err := s.neutron.ListNetworksV2()
+	c.Assert(err, gc.IsNil)
+	if len(networks) < 1 {
+		c.Errorf("at least one neutron network is necessary for this tests")
+	}
+	for _, network := range networks {
+		c.Check(network.Id, gc.Not(gc.Equals), "")
+		c.Check(network.Name, gc.Not(gc.Equals), "")
+	}
+	firstNetwork := networks[0]
+	foundNetwork, err := s.neutron.GetNetworkV2(firstNetwork.Id)
+	c.Assert(err, gc.IsNil)
+	c.Check(foundNetwork.Id, gc.Equals, firstNetwork.Id)
+	c.Check(foundNetwork.Name, gc.Equals, firstNetwork.Name)
+}
+
+func (s *LiveTests) TestSubnetsV2(c *gc.C) {
+	subnets, err := s.neutron.ListSubnetsV2()
+	c.Assert(err, gc.IsNil)
+	if len(subnets) < 1 {
+		c.Errorf("at least one neutron subnet is necessary for this tests")
+	}
+	for _, subnet := range subnets {
+		c.Check(subnet.Id, gc.Not(gc.Equals), "")
+		c.Check(subnet.NetworkId, gc.Not(gc.Equals), "")
+		c.Check(subnet.Name, gc.Not(gc.Equals), "")
+		_, _, err := net.ParseCIDR(subnet.Cidr)
+		c.Assert(err, gc.IsNil)
+	}
+	firstSubnet := subnets[0]
+	foundSubnet, err := s.neutron.GetSubnetV2(firstSubnet.Id)
+	c.Assert(err, gc.IsNil)
+	c.Check(foundSubnet.Id, gc.Equals, firstSubnet.Id)
+	c.Check(foundSubnet.NetworkId, gc.Equals, firstSubnet.NetworkId)
+	c.Check(foundSubnet.Name, gc.Equals, firstSubnet.Name)
+}
+
+func (s *LiveTests) deleteSecurityGroup(id string, c *gc.C) {
+	err := s.neutron.DeleteSecurityGroupV2(id)
+	c.Assert(err, gc.IsNil)
+}
+
+func (s *LiveTests) TestSecurityGroupsV2(c *gc.C) {
+	newSecGrp, err := s.neutron.CreateSecurityGroupV2("SecurityGroupTest", "Testing create security group")
+	c.Assert(err, gc.IsNil)
+	c.Assert(newSecGrp, gc.Not(gc.IsNil))
+	defer s.deleteSecurityGroup(newSecGrp.Id, c)
+	secGrps, err := s.neutron.ListSecurityGroupsV2()
+	c.Assert(err, gc.IsNil)
+	c.Assert(secGrps, gc.Not(gc.HasLen), 0)
+	var found bool
+	for _, secGrp := range secGrps {
+		c.Check(secGrp.Id, gc.Not(gc.Equals), "")
+		c.Check(secGrp.Name, gc.Not(gc.Equals), "")
+		c.Check(secGrp.Description, gc.Not(gc.Equals), "")
+		c.Check(secGrp.TenantId, gc.Not(gc.Equals), "")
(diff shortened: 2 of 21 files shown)


glance/glance.go
```diff
@@ -106,3 +106,57 @@ func (c *Client) GetImageDetail(imageId string) (*ImageDetail, error) {
 	}
 	return &resp.Image, nil
 }
+
+// ImageDetailV2 describes extended information about an image for API v2.0.
```
bz2: I know the upstream docs move around a  bit, but I'd like a url reference in the comment here so it's clearer what the various fields in the struct mean.

bz2, line 111: I know the upstream docs move around a bit, but I'd like a url reference in the comment here so it's clearer what the various fields in the struct mean.

neutron/local_test.go
```diff
@@ -0,0 +1,92 @@
...
+	s.LiveTests.SetUpSuite(c)
+}
+
+func (s *localLiveSuite) TearDownSuite(c *gc.C) {
+	s.LiveTests.TearDownSuite(c)
+	s.openstack.Stop()
+}
+
+func (s *localLiveSuite) SetUpTest(c *gc.C) {
+	s.retryErrorCount = 0
```
bz2: Why is setting this to zero required? Without this, can just omit SetUpTest and TearDownTest on localLiveSuite and LiveTests?

neutron/live_test.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+func (s *LiveTests) TestListNetworksV2(c *gc.C) {
+	networks, err := s.neutron.ListNetworksV2()
+	c.Assert(err, gc.IsNil)
+	for _, network := range networks {
+		c.Check(network.Id, gc.Not(gc.Equals), "")
+		c.Check(network.Name, gc.Not(gc.Equals), "")
+	}
+	firstNetwork := networks[0]
```
bz2, line 104: Test hasn't actually asserted this one or more networks returned? Is that just a given anyway?

neutron/live_test.go
```diff
@@ -0,0 +1,239 @@
...
+func (s *LiveTests) TestSubnetsV2(c *gc.C) {
+	subnets, err := s.neutron.ListSubnetsV2()
+	c.Assert(err, gc.IsNil)
+	for _, subnet := range subnets {
+		c.Check(subnet.Id, gc.Not(gc.Equals), "")
+		c.Check(subnet.NetworkId, gc.Not(gc.Equals), "")
+		c.Check(subnet.Name, gc.Not(gc.Equals), "")
+		c.Assert(subnet.Cidr, gc.Matches, `\d{1,3}(\.+\d{1,3}){3}\/\d+`)
+	}
+	firstSubnet := subnets[0]
```
bz2, line 124: Same issue, will there always be one ore more subnets?

neutron/live_test.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+func (s *LiveTests) TestSubnetsV2(c *gc.C) {
+	subnets, err := s.neutron.ListSubnetsV2()
+	c.Assert(err, gc.IsNil)
+	for _, subnet := range subnets {
+		c.Check(subnet.Id, gc.Not(gc.Equals), "")
+		c.Check(subnet.NetworkId, gc.Not(gc.Equals), "")
+		c.Check(subnet.Name, gc.Not(gc.Equals), "")
+		c.Assert(subnet.Cidr, gc.Matches, `\d{1,3}(\.+\d{1,3}){3}\/\d+`)
```
bz2: Are subnets always ipv4?

neutron/live_test.go
```diff
@@ -0,0 +1,239 @@
...
+		}
+	}
+	if netId == "" {
+		c.Errorf("no valid network to create floating IP")
+	}
+	c.Assert(netId, gc.Not(gc.Equals), "")
+	ip, err := s.neutron.AllocateFloatingIPV2(netId)
+	c.Assert(err, gc.IsNil)
+	c.Assert(ip, gc.Not(gc.IsNil))
+	c.Check(ip.IP, gc.Not(gc.Equals), "")
```
bz2, line 61: If any of these following checks fails, the floating ip remains allocated?

neutron/live_test.go
```diff
@@ -0,0 +1,239 @@
...
+	c.Check(foundSubnet.NetworkId, gc.Equals, firstSubnet.NetworkId)
+	c.Check(foundSubnet.Name, gc.Equals, firstSubnet.Name)
+	c.Assert(foundSubnet.Cidr, gc.Matches, `\d{1,3}(\.+\d{1,3}){3}\/\d+`)
+}
+
+func (s *LiveTests) TestSecurityGroupsV2(c *gc.C) {
+	newSecGrp, err := s.neutron.CreateSecurityGroupV2("SecurityGroupTest", "Testing create security group")
+	c.Assert(err, gc.IsNil)
+	c.Assert(newSecGrp, gc.Not(gc.IsNil))
+	defer s.neutron.DeleteSecurityGroupV2(newSecGrp.Id)
```
bz2: This should really be a function that also checks the error return from DeleteSecurityGroupV2.

neutron/live_test.go
```diff
@@ -0,0 +1,239 @@
...
+	c.Assert(err, gc.Not(gc.IsNil))
+	err = s.neutron.DeleteSecurityGroupV2(updatedSecGroup.Id)
+	c.Assert(err, gc.IsNil)
+}
+
+func (s *LiveTests) TestSecurityGroupsByNameV2(c *gc.C) {
+	// Create and find a SecurityGroup
+	newSecGrp, err := s.neutron.CreateSecurityGroupV2("SecurityGroupTest", "Testing find security group by name")
+	c.Assert(err, gc.IsNil)
+	defer s.neutron.DeleteSecurityGroupV2(newSecGrp.Id)
```
bz2: Same thing with err being unchecked.

neutron/live_test.go
```diff
@@ -0,0 +1,239 @@
...
+	c.Assert(newSecGrp, gc.Not(gc.IsNil))
+	foundSecGrps, err := s.neutron.SecurityGroupByNameV2(newSecGrp.Name)
+	c.Assert(err, gc.IsNil)
+	c.Assert(foundSecGrps, gc.HasLen, 1)
+	if newSecGrp.Id != foundSecGrps[0].Id {
+		c.Errorf("expected to find added security group %s, when requested by name", newSecGrp.Name)
+	}
+	// Try to find a SecurityGroup that doesn't exist
+	errorSecGrps, err := s.neutron.SecurityGroupByNameV2("deadbeef")
```
bz2: I agree that's an unlike name, but I'd generally use something like 'nonexisistantgroup' so any error messages are clearer.

neutron/live_test.go
```diff
@@ -0,0 +1,239 @@
...
+	c.Assert(err, gc.Not(gc.IsNil))
+	c.Assert(errorSecGrps, gc.HasLen, 0)
+	// Create and find a SecurityGroup with spaces in the name
```
bz2, line 195: This could just be a seperate test.

neutron/local_test.go
```diff
@@ -0,0 +1,92 @@
...
+	s.LiveTests.SetUpTest(c)
+}
+
+func (s *localLiveSuite) TearDownTest(c *gc.C) {
+	s.LiveTests.TearDownTest(c)
+}
+
+// Additional tests to be run against the service double only go here.
```
bz2: This comment is not a doc comment on the method below. Delete it?

neutron/neutron.go
```diff
@@ -0,0 +1,381 @@
...
+	ApiNetworksV2           = "networks"
+	ApiSubnetsV2            = "subnets"
+	ApiSecurityGroupsV2     = "security-groups"
+	ApiSecurityGroupRulesV2 = "security-group-rules"
+)
+
+// NetworkV2 contains details about a labeled network
+type NetworkV2 struct {
+	Id                string   `json:"id"` // UUID of the resource
+	Name              string   // User-provided name for the network range
```
bz2, line 27: Doesn't this need `json:"name"` as well?

neutron/neutron.go
```diff
@@ -0,0 +1,381 @@
...
+	External          bool     `json:"router:external"` // is this network connected to an external router
+	AvailabilityZones []string `json:"availability_zones"`
+	TenantId          string   `json:"tenant_id"`
+}
+
+// SubnetV2 contains details about a labeled subnet
+type SubnetV2 struct {
+	Id              string        `json:"id"`         // UUID of the resource
+	NetworkId       string        `json:"network_id"` // UUID of the related network
+	Name            string        // User-provided name for the subnet
```
bz2: Also needs `json:"name"`?

neutron/neutron.go
```diff
@@ -0,0 +1,381 @@
...
+// OpenStack now supports filtering with API calls.
+// More than one Security Group may be returned, as names are not unique
+// e.g. name=default
+func (c *Client) SecurityGroupByNameV2(name string) ([]SecurityGroupV2, error) {
+	var resp struct {
+		Groups []SecurityGroupV2 `json:"security_groups"`
+	}
+	// Can't have any spaces in a url
+	noSpaceName := name
+	if strings.Contains(name, " ") {
```
bz2: Can't we do proper url encoding instead?

neutron/neutron.go
```diff
@@ -0,0 +1,381 @@
...
+	Direction      string  `json:"direction"` // Required
+	Id             string  `json:",omitempty"`
+	TenantId       string  `json:"tenant_id,omitempty"`
+}
+
+// SecurityGroupV2 describes a single security group in OpenStack.
+type SecurityGroupV2 struct {
+	Rules       []SecurityGroupRuleV2 `json:"security_group_rules"`
+	TenantId    string                `json:"tenant_id"`
+	Id          string
```
bz2: I'd give json lowercased names for these as well so they reserialise lowercased.

neutron/neutron.go
```diff
@@ -0,0 +1,381 @@
...
+	if len(resp.Groups) == 0 {
+		return nil, errors.Newf(err, "failed to find security group with name: %s", name)
+	}
+	return resp.Groups, nil
+}
+
+// CreateSecurityGroupV2 creates a new security group.
+func (c *Client) CreateSecurityGroupV2(name, description string) (*SecurityGroupV2, error) {
+	var req struct {
+		SecurityGroupV2 struct {
```
bz2, line 240: With omitempty on the rules, tenantid, and id, could use the normal SecurityGroupV2 struct.

neutron/neutron.go
```diff
@@ -0,0 +1,381 @@
...
+	url := fmt.Sprintf("%s/%s", ApiSecurityGroupsV2, groupId)
+	requestData := goosehttp.RequestData{ReqValue: req, RespValue: &resp, ExpectedStatus: []int{http.StatusOK}}
+	err := c.client.SendRequest(client.PUT, "network", "v2.0", url, &requestData)
+	if err != nil {
+		return nil, errors.Newf(err, "failed to update security group with Id %s to name: %s", groupId, name)
+	}
+	return &resp.SecurityGroup, nil
+}
+
+// RuleInfo allows the callers of CreateSecurityGroupRuleV2() to
```
bz2: RuleInfoV2.

testservices/neutronmodel/neutronmodel.go
```diff
@@ -0,0 +1,616 @@
...
+			SubnetIds:         []string{"998-01"},
+			External:          true,
+			AvailabilityZones: []string{"test-available"},
+		},
+		{
+			Id:                "997",
+			Name:              "ext-net-wrong-az",
+			SubnetIds:         []string{"997-01"},
+			External:          true,
+			AvailabilityZones: []string{"deadbeef"},
```
bz2: I'd use a string like 'another-az' rather than the deadbeef.

testservices/neutronmodel/neutronmodel.go
```diff
@@ -0,0 +1,616 @@
...
+func (n *NeutronModel) AddSecurityGroup(group neutron.SecurityGroupV2) error {
+	n.rwMu.Lock()
+	defer n.rwMu.Unlock()
+	if _, err := n.SecurityGroup(group.Id); err == nil {
+		return testservices.NewSecurityGroupAlreadyExistsError(group.Id)
+	}
+	// Neutron adds 2 default egress security group rules to new security
+	// groups, copy that behavior here.
+	id, _ := strconv.Atoi(group.Id)
+	id1 := id * 999
```
bz2, line 172: Multiply rather than mask/add for unique ids is interesting, test output might get some very large ints. :)

testservices/neutronservice/service_http.go
```diff
@@ -0,0 +1,719 @@
...
+	}
+	errBadRequest = &errorResponse{
+		http.StatusBadRequest,
+		`{"badRequest": {"message": "Malformed request url", "code": 400}}`,
+		"application/json; charset=UTF-8",
+		"bad request base path or URL",
+		nil,
+		nil,
+	}
+	errBadRequest2 = &errorResponse{
```
bz2: I'd try and give these clearer names, eg errBadRequestMalformedURL/errBadRequestIncorrect/errBadRequestMalformedBody.

testservices/neutronservice/service_http.go
```diff
@@ -0,0 +1,719 @@
...
+		http.StatusOK,
+		`{"versions": [{"status": "CURRENT", "id": "v2.0", "links": [{"href": "v2.0", "rel": "self"}]}]}`,
+		"application/json",
+		"no version specified in URL",
+		nil,
+		nil,
+	}
+	errVersionsLinks = &errorResponse{
+		http.StatusOK,
+		`{"version": {"status": "CURRENT", "updated": "2011-01-21T11` +
```
bz2: Well, isn't this a nice error.

testservices/novaservice/service.go
```diff
@@ -801,27 +880,79 @@ func (n *Nova) addServerFloatingIP(serverId string, ipId string) error {
...
+
+// addFloatingIPToServerAddresses adds a floating ip address to the servers list
+// of Addresses to facilitate juju openstack provider tests.
+func (n *Nova) addFloatingIPToServerAddresses(serverId, address string) error {
+	server, err := n.server(serverId)
+	if err != nil {
+		return err
+	}
+	newAddresses := server.Addresses["private"]
+	if strings.Contains(address, ":") {
```
bz2, line 897: Could use net.IP to do a more sane determination of 4 vs 6, but doesn't matter much for test code.

testservices/novaservice/service_http_test.go
```diff
@@ -18,30 +18,39 @@ import (
...
 	httpsuite.HTTPSuite
-	service *Nova
-	token   string
+	service              *Nova
+	token                string
+	useNeutronNetworking bool
 }
 
-var _ = gc.Suite(&NovaHTTPSSuite{HTTPSuite: httpsuite.HTTPSuite{UseTLS: true}})
+var _ = gc.Suite(&NovaHTTPSSuite{HTTPSuite: httpsuite.HTTPSuite{UseTLS: true}, useNeutronNetworking: true})
```
bz2, line 40: It's an arbitrary pick that the https varient uses the new codepaths and the other doesn't? I suspect that doing a matrix of these gets us very little though with how the testing is structured.

testservices/novaservice/service_test.go
```diff
@@ -21,10 +23,17 @@ const (
 	region      = "region"
 )
 
-var _ = gc.Suite(&NovaSuite{})
+var _ = gc.Suite(&NovaSuite{useNeutronNetworking: true})
+var _ = gc.Suite(&NovaSuite{useNeutronNetworking: false})
```
bz2, line 27: Okay, so these tests just do both paths.

bz2 (changes requested):
Thanks for this!

I have a bunch of comments inline, not all of them are actionable changes, please poke me if anything is unclear or if you disagree, am happy to discuss design questions.

Skimmed the test service chunks a bit faster than the rest of the code, and yes, I do feel your pain here. Attempting to have a setup that sort-of runs 'live' tests rather than more focused unit testing leads to a crazy amount of test code to support a small quantity of business logic.

Bigger questions on the PR overall:
- What versions of openstack are expecting the neutron code to work with... L upwards? What did you test against when developing?
- How is a user of the goose library expected to select between nova networking and neutron? In other words, if they just want to make a security group, what should their code do to select the right set of functions for their openstack? Or, do we just think everyone should be using neutron regardless these days?

I am mgz on irc/freenode if you want to poke me in real time.

hmlanigan:
We're expecting this code to work with IceHouse and forward, until the APIs are deprecated or new ones revealed as better.  Primary testing has been done against Mitaka.  And a tiny bit against Ice House.  Ice House is the earliest OpenStack release supported with juju 2.0.

Ideally all OpenStack deployments are using Neutron networking at this point.  However I can't guarantee that.  Nor do I fully know which other projects are using go-goose.  From a package perspective I chose to leave the Nova networking in place as the code doesn't conflict.   

Developers using using goose will have to choose the Neutron functions rather than the Nova functions if they want to use Neutron networking.  The function names are very similar, so should be easy to find.  Functions within the Nova package which aren't directly related to creating or listing floating ips, security groups and rules, or networks continue to function independently of Neutron.  Such as nova.AddServerFloatingIP().

However if a developer is using the OpenStack test service, they will be using neutron networking and the Nova networking functions won't work.  If they use the Nova test service, a choice can be made on whether to use nova or neutron.

glance/glance.go
hmlanigan: The url is at the top of the page.  I'm adding one for the image functions utilizing the compute server as well.

neutron/local_test.go
hmlanigan: You've caught me doing a copy, paste, modify from the nova tests to create the neutron ones and not catching all of the clean up work.  I removed code related to retryErrorCount and retryErrorCountToSend.

neutron/local_test.go
hmlanigan: yes

neutron/live_test.go
hmlanigan: added defer DeleteFloatingIPV2()

neutron/live_test.go
hmlanigan: For the test doubles, there will always be a network.  If testing live, we need an existing network to run the test, as we don't create any.  Adding an error if no networks exist.

neutron/live_test.go
hmlanigan: same answer as for networks

neutron/live_test.go
hmlanigan: No. IPv4 is just what is used by the code I'm testing against.  Updating to use net.ParseCidr().  Adding IPv6 subnet to test double.

neutron/live_test.go
hmlanigan: done.

neutron/live_test.go
hmlanigan: most of the tests could be broken up into multiple tests.  i'm going to leave this as is.

neutron/neutron.go
hmlanigan: the json parsing seems to be okay converting if the capitalization is different between the structures.  However the code isn't consistent about when to specify json:"", so I'll add it to all.

neutron/neutron.go
hmlanigan: see above

neutron/neutron.go
hmlanigan: A rules array isn't optional when a SecurityGroup is returned from neutron, trying to stay consistent with the API.

neutron/neutron.go
hmlanigan: done

testservices/neutronmodel/neutronmodel.go
hmlanigan: resolved

testservices/neutronmodel/neutronmodel.go
hmlanigan: luckily i wrote the test code.  :-)

testservices/neutronservice/service_http.go
hmlanigan: and isn't used... removing.

neutron/neutron.go
hmlanigan: found url.QueryEscape(), using that instead.

testservices/novaservice/service_http_test.go
hmlanigan: Yes, it is an arbitrary pick, however I saw no reason to double the number of tests to complete the matrix.  The two code paths are unrelated.

testservices/neutronservice/service_http.go
hmlanigan: done

neutron/live_test.go
hmlanigan: done

neutron/live_test.go
hmlanigan: done

bz2 (approved):
Thanks for making the changes, I think this is good to go now.

bz2:
$$merge$$

PR #31, opened 2016-11-23
Handle 404 in version discovery gracefully

axw:
If, when attempting API version discovery, we receive
a 404 response when querying the root endpoint, we
should not prevent the API request. Instead, revert
to the old behaviour of using the service catalogue's
public URL entry as-is.

### DIFF
--- a/client/apiversion.go
+++ b/client/apiversion.go
@@ -9,6 +9,7 @@ import (
 	"strconv"
 	"strings"
 
+	gooseerrors "gopkg.in/goose.v1/errors"
 	goosehttp "gopkg.in/goose.v1/http"
 )
 
@@ -134,7 +135,10 @@ func unmarshallVersion(Versions json.RawMessage) ([]apiVersionInfo, error) {
 	return versions, nil
 }
 
-// getAPIVersions returns data on the API versions supported by the specified service endpoint.
+// getAPIVersions returns data on the API versions supported by the specified
+// service endpoint. Some OpenStack clouds do not support the version endpoint,
+// in which case this method will return an empty set of versions in the result
+// structure.
 func (c *authenticatingClient) getAPIVersions(serviceCatalogURL string) (*apiURLVersion, error) {
 	url, err := url.Parse(serviceCatalogURL)
 	if err != nil {
@@ -179,7 +183,7 @@ func (c *authenticatingClient) getAPIVersions(serviceCatalogURL string) (*apiURL
 		return &apiURLVersion{*url, strings.Join(origPathParts, "/"), objectStoreApiVersionInfo}, nil
 	}
 
-	// make sure we haven't already received the version info
+	// Make sure we haven't already received the version info.
 	c.apiVersionMu.Lock()
 	defer c.apiVersionMu.Unlock()
 	if apiInfo, ok := c.apiURLVersions[baseURL]; ok {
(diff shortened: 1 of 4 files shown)


axw:
**QA**

- bootstrapped neutron
- bootstrapped canonistack

client/apiversion.go
```diff
@@ -196,18 +200,29 @@ func (c *authenticatingClient) getAPIVersions(serviceCatalogURL string) (*apiURL
 			http.StatusMultipleChoices,
 		},
 	}
+	apiURLVersionInfo := &apiURLVersion{
+		rootURL:          *url,
+		serviceURLSuffix: strings.Join(pathParts, "/"),
+	}
 	if err := c.sendRequest("GET", baseURL, c.Token(), requestData); err != nil {
+		if gooseerrors.IsNotFound(err) {
+			// Some OpenStack clouds do not support the
```
bz2, line 209: Rather than comment, can we trace log here instead? Urk, actually we're still on the stdlib logging package so there's no default suppressed log method.

bz2 (approved):
Okay, checking I understand this change correctly.

User supplies http://example.com/v2.0 endpoint. Then getAPIVersions parses it, finds the version, chops that off. Then tries to get the version index at http://example.com/ which in this case the cloud doesn't supply (maybe just not serving root url). The correct behaviour should be to just go ahead and use the url/version given.

So, I guess my only big picture question is whether it's better to return an empty list of versions and have the caller fall back to using the original serviceURL as implemented, or to return a list of just the version that was in the url. The logic seems somewhat fragile to finding random version-like things in url paths as evidenced by the rackspace special case.

client/apiversion.go
axw: > Okay, checking I understand this change correctly.

Almost, but with enough of a twist to explain the approach.

> User supplies http://example.com/v2.0 endpoint. Then getAPIVersions parses it, finds the version, chops that off. Then tries to get the version index at http://example.com/ which in this case the cloud doesn't supply (maybe just not serving root url). The correct behaviour should be to just go ahead and use the url/version given.

User supplies the URL of *keystone*, which is http://keystone.example.com/v2.0. We authenticate against this URL (there may also be some version discovery going on there, but it's irrelevant - auth succeeds). In the auth response, you get back the service catalogue.

The service catalogue contains services with types like "volume", "network", "compute", and each one has a "public URL". The public URL may or may not have a version included in the URL, and may or may not have some other junk like a tenant ID, e.g.

    compute: http://nova.example.com/v2/tenant
    network: http://neutron.example.com/

We will query the root URL (nova.example.com or neutron.example.com), and it should respond with the versions supported in one of two standard formats. We'll then choose the appropriate version from that list based on what the caller asked for, and form the URL according to the rel link in the entry. That link will include the version, so we'll end up with something like:

    http://nova.example.com/v2.1

(and then we add back the junk-like-tenant-ID on the end, so you get /v2.1/tenant)

> So, I guess my only big picture question is whether it's better to return an empty list of versions and have the caller fall back to using the original serviceURL as implemented, or to return a list of just the version that was in the url.

So that's what we end up choosing. The reason why we don't include the URL in the list, is because we don't have a version to associate it with. In some cases the service catalogue's public URL has a version, and sometimes not; so we can't necessarily infer the version from the URL.

> The logic seems somewhat fragile to finding random version-like things in url paths as evidenced by the rackspace special case.

Not sure, but I think those comments about rackspace are not quite right. Swift is special, and does not support the version endpoint, regardless of which OpenStack cloud.

axw: > Rather than comment, can we trace log here instead? Urk, actually we're still on the stdlib logging package so there's no default suppressed log method.

Yes, I would like to pass in a logger interface to the client, and use it. I opened a bug about doing that just recently. Would you be OK with me doing that in a follow up?

bz2: Absolutely. And thank you for the longer explanation.

axw:
$$merge$$

PR #32, opened 2016-11-23
Fix Neutron test service duplicate rule check

fnordahl:
Current check erronously marks two rules with different RemoteIPPrefix
as duplicates.

Required by: juju/juju#6563

### DIFF
--- a/testservices/neutronservice/service_http_test.go
+++ b/testservices/neutronservice/service_http_test.go
@@ -517,6 +517,14 @@ func (s *NeutronHTTPSuite) TestAddSecurityGroupRule(c *gc.C) {
 		IPProtocol:     "tcp",
 		RemoteIPPrefix: "1.2.3.4/5",
 	}
+	riIngress2 := neutron.RuleInfoV2{
+		ParentGroupId:  "1",
+		Direction:      "ingress",
+		PortRangeMax:   22,
+		PortRangeMin:   22,
+		IPProtocol:     "tcp",
+		RemoteIPPrefix: "2.3.4.5/6",
+	}
 	riEgress := neutron.RuleInfoV2{
 		ParentGroupId:  group2.Id,
 		Direction:      "egress",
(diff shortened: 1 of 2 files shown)


testservices/neutronservice/service_http_test.go
```diff
@@ -562,15 +580,33 @@ func (s *NeutronHTTPSuite) TestAddSecurityGroupRule(c *gc.C) {
...
 	assertJSON(c, resp, &expected)
 	c.Assert(expected.Rule.Id, gc.Equals, rule2.Id)
 	c.Assert(expected.Rule.ParentGroupId, gc.Equals, rule2.ParentGroupId)
-	err = s.service.removeSecurityGroupRule(rule2.Id)
+	c.Assert(*expected.Rule.PortRangeMax, gc.Equals, *rule2.PortRangeMax)
+	c.Assert(*expected.Rule.PortRangeMin, gc.Equals, *rule2.PortRangeMin)
+	c.Assert(*expected.Rule.IPProtocol, gc.Equals, *rule2.IPProtocol)
+	c.Assert(expected.Rule.Direction, gc.Equals, rule2.Direction)
+	c.Assert(expected.Rule.RemoteIPPrefix, gc.Equals, rule2.RemoteIPPrefix)
+	defer s.service.removeSecurityGroupRule(rule2.Id)
```
bz2, line 601: Really this defer cleanup should go immediately after the `(err, gc.IsNil)` `(resp.StatusCode, gc.Equals, http.StatusCreated)` lines and before the content assertions, and also check its own error.

However, as this is just testing the test service, this kind of cleanup isn't strictly required.

bz2 (approved):
Thanks!

bz2:
$$merge$$

testservices/neutronservice/service_http_test.go
fnordahl: Thx! Will address these in the next PR  currently under review with a updated patch set before merge.

PR #33, opened 2016-11-24
Add CompatLogger interface, loggo impl

axw:
We change several functions to take a new
CompatLogger instead of *log.Logger, and
provide a way of upgrading the interface
to a loggo.Logger.

Technically this is an API break, because
the signature of several exported functions
have changed. The CompatLogger interface is
implemented by *log.Logger, so if you're
just calling the functions and not taking
the function values, you'll be fine.


logging/logger.go
```diff
@@ -0,0 +1,31 @@
...
+// Licensed under the LGPLv3, see COPYING and COPYING.LESSER file for details.
+
+package logging
+
+import "github.com/juju/loggo"
+
+// CompatLogger is a minimal logging interface that may be provided
+// when constructing a goose Client to log requests and responses,
+// retaining compatibility with the old *log.Logger that was
+// previously dependend upon directly.
```
bz2: Typo, *depended

bz2 (commented):
Changing signatures to take a log.Logger compatible interface is I think an api break we can live with.

What's the reasoning behind having the unwrapping/extending local to each package, rather than exported in the logging package? The implementations are nice and small, but also mostly the same.

axw:
> What's the reasoning behind having the unwrapping/extending local to each package, rather than exported in the logging package? The implementations are nice and small, but also mostly the same.

Not really sure what I was thinking TBH. I've updated the logging package, dropped the internal interfaces/unwrapping code.

Also fixed unwrapping to check for nil *log.Logger, since it's converted to a non-nil iface value.

bz2 (approved):
Great, thanks Andrew!

axw:
$$merge$$

bz2:
$$merge$$

PR #36, opened 2016-11-29
Add additional potential domain env vars

Jamstah:
Potential variables for the domain are listed in this openstack doc: http://docs.openstack.org/developer/python-openstackclient/authentication.html

### DIFF
--- a/identity/identity.go
+++ b/identity/identity.go
@@ -120,9 +120,12 @@ var (
 		"OS_TENANT_NAME",
 		"NOVA_PROJECT_ID",
 	}
-	// CredEnvDomainName is used for Credentials.TenantName.
+	// CredEnvDomainName is used for Credentials.DomainName.
 	CredEnvDomainName = []string{
+		"OS_USER_DOMAIN_NAME",
+		"OS_PROJECT_DOMAIN_NAME",
 		"OS_DOMAIN_NAME",
+		"OS_DEFAULT_DOMAIN",
 	}
 )
 


Jamstah:
@axw would you mind reviewing this one?

identity/identity.go
```diff
@@ -120,9 +120,12 @@ var (
 		"OS_TENANT_NAME",
 		"NOVA_PROJECT_ID",
 	}
-	// CredEnvDomainName is used for Credentials.TenantName.
+	// CredEnvDomainName is used for Credentials.DomainName.
 	CredEnvDomainName = []string{
+		"OS_USER_DOMAIN_NAME",
+		"OS_PROJECT_DOMAIN_NAME",
 		"OS_DOMAIN_NAME",
+		"OS_DEFAULT_DOMAIN",
```
mjs, line 128: I know very little about OpenStack. Can you explain why you chose these env vars specifically? There's more listed here: http://docs.openstack.org/developer/python-openstackclient/authentication.html#authenticating-using-identity-server-api-v3

mjs (commented):
This is probably fine but I just want some clarification.

Jamstah:
Do you mean the _ID ones? Could be a good idea to add them but I don't know goose so it would need testing to check it still works, or if it would need extending to cope.

I choose them based on the open stack doc and my experience, and guessed at a sensible order. I haven't checked which ones the open stack command line reads, or their preference order, that could be another thing to do.

What do you think?

mjs:
Sounds OK to me. Ship it :)

mjs (approved):

wallyworld:
Closing this PR as support for the extra env vars was recently added in a different PR.

PR #37, opened 2016-11-30
Remove Content-Type header for payloadless request

gnuoy:
If a call out to the OpenStack API does not contain a
payload then do not set a Content-Type header as this
causes the Request to fail for OpenStack Mitaka+
deployments. Fixes lp:1625624

QA Steps
-------------

To reproduce the problem run the steps below. The final secgroup-list should not return
any security groups for this environment if Juju has cleaned up properly.

```
$ juju bootstrap
$ nova secgroup-list | grep jujutest     
$ time juju destroy-environment --yes jujutest                          
$ nova secgroup-list | grep jujutest     
```

### DIFF
--- a/http/client.go
+++ b/http/client.go
@@ -114,7 +114,7 @@ func gooseAgent() string {
 	return fmt.Sprintf("goose (%s)", goose.Version)
 }
 
-func createHeaders(extraHeaders http.Header, contentType, authToken string) http.Header {
+func createHeaders(extraHeaders http.Header, contentType, authToken string, payloadExists bool) http.Header {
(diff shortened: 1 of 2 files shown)


bz2 (approved):
Looks good, thanks!

bz2:
$$merge$$

bz2:
$$retry$$

```
--- FAIL: TestRunSuccess (0.00s)
FAIL
FAIL	gopkg.in/goose.v1/sync	0.020s
```

That... is not something this branch in any way touches, perhaps the 1ms timeout is too optimistic given load.

PR #41, opened 2017-04-03
Fix for Bug 38: Adding optional filter to Neutron ListNetworksV2()

hmlanigan:
Filter for Router:External and Name available currently.  There should be no impact to current
ListNetworksV2() calls.

### DIFF
--- a/neutron/neutron.go
+++ b/neutron/neutron.go
@@ -21,6 +21,13 @@ const (
 	ApiSecurityGroupRulesV2 = "security-group-rules"
 )
 
+// Filter keys for Networks.
+// As of the Newton release of OpenStack, Network filter by subnet was not implemented
+const (
+	FilterRouterExternal    = "router:external"    // The router:external
+	FilterNetwork           = "name"               // The network name.
+)
+
 // NetworkV2 contains details about a labeled network
 type NetworkV2 struct {
 	Id                string   `json:"id"` // UUID of the resource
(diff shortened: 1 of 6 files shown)


neutron/neutron.go
```diff
@@ -51,12 +58,45 @@ func New(client client.Client) *Client {
...
 
-// ListNetworksV2 gives details on available networks
-func (c *Client) ListNetworksV2() ([]NetworkV2, error) {
+// ----------------------------------------------------------------------------
+// TODO(hml): copied from the nova package.  However it should really be pulled out
+// and shared between goose pkgs, however we don't want to break compatibility at this time.
+//
+// Filtering helper.
+//
+// Filter builds filtering parameters to be used in an OpenStack query which supports
```
axw: please move the "Filter builds ..." to the top, so the docs lead with that, per standard go docs practices

neutron/neutron.go
```diff
@@ -51,12 +58,45 @@ func New(client client.Client) *Client {
...
+	return &Filter{make(url.Values)}
+}
+
+func (f *Filter) Set(filter, value string) {
+	f.v.Set(filter, value)
+}
+
+// ----------------------------------------------------------------------------
+
+// ListNetworksV2 gives details on available networks, zero or one Filters
```
axw, line 89: why not just have it take a `*Filter`, which may be nil? that would make it the same as other methods in goose

is this to avoid bumping the version? TBH I think it might be time anyway. is it to onerous to bump to say, v2-unstable, and then we can iron out the kinks over a period of time and then call it v2 (but still use v2-unstable in juju along the way)?

axw (commented):
Looks good, except for the surprising method signature.

neutron/neutron.go
hmlanigan: Yes, at the moment, we're trying avoid bumping the goose version to get Bug 1654144 fixed in juju 2.2.  The plan is to bump to to v2 in the next month or so, to go with other changes needed for the next release of juju.

neutron/neutron.go
hmlanigan: done.

neutron/neutron.go
axw: OK. Thanks for adding the TODO.

axw (approved):

hmlanigan:
$$merge$$

axw:
$$merge$$

PR #43, opened 2017-04-19
Fix Bug 23 nova live test:

hmlanigan:
* Remove requirement for Swift OpenStack Service
* Add network flag to live config to specify uuid of network for RunServer
* Add netType flag to live config, nova or neutron allowed
* Skip tests which are nova network specific if using neutron during live tests
* Add waitForTestServerToStart so test is run correctly

### DIFF
--- a/nova/live_test.go
+++ b/nova/live_test.go
@@ -20,12 +20,14 @@ const (
 	testImageName = "nova_test_server"
 )
 
-func registerOpenStackTests(cred *identity.Credentials, testImageDetails imageDetails) {
+func registerOpenStackTests(cred *identity.Credentials, testDetails instanceDetails) {
 	gc.Suite(&LiveTests{
-		cred:        cred,
-		testImageId: testImageDetails.imageId,
-		testFlavor:  testImageDetails.flavor,
-		vendor:      testImageDetails.vendor,
+		cred:                 cred,
+		testImageId:          testDetails.imageId,
+		testFlavor:           testDetails.flavor,
+		testNetwork:          testDetails.network,
+		vendor:               testDetails.vendor,
+		useNeutronNetworking: testDetails.useNeutronNetworking,
 	})
 }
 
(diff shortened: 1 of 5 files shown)


nova/nova_test.go
```diff
@@ -3,44 +3,49 @@ package nova_test
...
 	"gopkg.in/goose.v2/identity"
 )
 
-// imageDetails specify parameters used to start a test machine for the live tests.
-type imageDetails struct {
-	flavor  string
-	imageId string
-	vendor  string
+// details specify parameters used to start a test machine for the live tests.
+type details struct {
```
wallyworld: maybe instanceDetails is a better name than just details

nova/nova_test.go
```diff
@@ -3,44 +3,49 @@ package nova_test
...
+	useNeutronNetworking bool
+	flavor               string
+	imageId              string
+	network              string
+	vendor               string
 }
 
 // Out-of-the-box, we support live testing using Canonistack or HP Cloud.
```
wallyworld: this comment can be removed i think

nova/nova_test.go
```diff
@@ -50,13 +55,25 @@ func Test(t *testing.T) {
...
+				t.Fatalf("Must specify network by UUID to use for test instance, "+
+					"eg %s for Canonistack", "-network 4e408154-e71a-4702-a8fd-edb8df1b4e6c")
+			}
+			var useNeutronNetworking bool
+			if strings.ToLower(*netType) == "neutron" {
+				useNeutronNetworking = true
+			} else if strings.ToLower(*netType) == "nova" {
+				useNeutronNetworking = false
+			} else {
+				t.Fatalf("netType must be Neutron or Nova")
```
wallyworld: the message should just use lower case for the names

nova/live_test.go
```diff
@@ -373,11 +397,17 @@ func (s *LiveTests) TestGetServer(c *gc.C) {
 	s.assertServerDetails(c, server)
 }
 
-func (s *LiveTests) waitTestServerToStart(c *gc.C) {
+func (s *LiveTests) waitTestServerToStart(c *gc.C, id ...string) {
 	// Wait until the test server is actually running
-	c.Logf("waiting the test server %s to start...", s.testServer.Id)
+	var serverId string
```
wallyworld: I don't quite get why we are using a slice here for id
It would be better just to take an explicit id arg and wait on that.
Callers would always pass in an id - either a bespoke id or s.testServer.Id

The above is much simpler to implement and understand, even though it means changing a few callers to pass in s.testServer.Id

wallyworld (approved):
A few small things to fix

nova/nova_test.go
hmlanigan: updated to remove HP Cloud

nova/nova_test.go
hmlanigan: done

nova/nova_test.go
hmlanigan: done

nova/live_test.go
hmlanigan: done

hmlanigan:
$$merge$$

PR #44, opened 2017-04-27
Add PortSecurityEnabled to NetworksV2. 

hmlanigan:
... Update testservices to not add
SecurityGroups if PortSecurityEnabled is false and add server SecurityGroups
to the ServerDetail returned.

NeutronFilter should not be a regexp to match real life behavior.  Update tests accordingly.

This is required to fix juju openstack provider bug: https://bugs.launchpad.net/juju/+bug/1680787.

Tested with test service and with juju against openstack networks with port_security_enabled set to true and false.

### DIFF
--- a/neutron/neutron.go
+++ b/neutron/neutron.go
@@ -30,12 +30,13 @@ const (
 
 // NetworkV2 contains details about a labeled network
 type NetworkV2 struct {
-	Id                string   `json:"id"` // UUID of the resource
-	Name              string   // User-provided name for the network range
-	SubnetIds         []string `json:"subnets"`         // an array of subnet UUIDs
-	External          bool     `json:"router:external"` // is this network connected to an external router
-	AvailabilityZones []string `json:"availability_zones"`
-	TenantId          string   `json:"tenant_id"`
+	Id                  string   `json:"id"` // UUID of the resource
+	Name                string   // User-provided name for the network range
+	SubnetIds           []string `json:"subnets"`         // an array of subnet UUIDs
+	External            bool     `json:"router:external"` // is this network connected to an external router
+	AvailabilityZones   []string `json:"availability_zones"`
+	TenantId            string   `json:"tenant_id"`
+	PortSecurityEnabled *bool    `json:"port_security_enabled"`
 }
 
 // SubnetV2 contains details about a labeled subnet
--- a/nova/nova.go
+++ b/nova/nova.go
@@ -232,8 +232,7 @@ type ServerDetail struct {
 	Name     string
 	Metadata map[string]string
 
-	// HP Cloud returns security groups in server details.
-	Groups []Entity `json:"security_groups"`
+	Groups *[]SecurityGroupName `json:"security_groups"`
 
 	// Progress holds the completion percentage of
 	// the current operation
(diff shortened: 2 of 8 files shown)


nova/nova.go
```diff
@@ -437,12 +436,9 @@ func (c *Client) GetServerSecurityGroups(serverId string) ([]SecurityGroup, erro
 		if errors.IsNotFound(err) {
 			serverDetails, err := c.GetServer(serverId)
 			if err == nil {
-				result := make([]SecurityGroup, len(serverDetails.Groups))
-				for i, e := range serverDetails.Groups {
-					result[i] = SecurityGroup{
-						Id:   e.Id,
-						Name: e.Name,
-					}
+				result := make([]SecurityGroup, len(*serverDetails.Groups))
```
babbageclunk: Do we need to check for `serverDetails.Groups` being nil here?

neutron/neutron.go
```diff
@@ -30,12 +30,13 @@ const (
...
-	External          bool     `json:"router:external"` // is this network connected to an external router
-	AvailabilityZones []string `json:"availability_zones"`
-	TenantId          string   `json:"tenant_id"`
+	Id                  string   `json:"id"` // UUID of the resource
+	Name                string   // User-provided name for the network range
+	SubnetIds           []string `json:"subnets"`         // an array of subnet UUIDs
+	External            bool     `json:"router:external"` // is this network connected to an external router
+	AvailabilityZones   []string `json:"availability_zones"`
+	TenantId            string   `json:"tenant_id"`
+	PortSecurityEnabled *bool    `json:"port_security_enabled"`
```
babbageclunk, line 39: Why a `*bool` here rather than a `bool`? (I guess you need for it to be nil-able?) Should `External` be `*bool` too?

testservices/novaservice/service.go
```diff
@@ -304,6 +305,16 @@ func (n *Nova) server(serverId string) (*nova.ServerDetail, error) {
 	if !ok {
 		return nil, testservices.NewServerByIDNotFoundError(serverId)
 	}
+	groups := n.allServerSecurityGroups(serverId)
+	if len(groups) > 0 {
+		groupNames := make([]nova.SecurityGroupName, 0)
```
babbageclunk: You can leave out the 0 here. Alternatively if you want to avoid growing the slice you can specify the length of groups:
```go
groupNames := make([]nova.SecurityGroupName, len(groups))
for i, group := range groups {
    groupNames[i] = nova.SecurityGroupName{Name: group.Name})
}
```

babbageclunk (approved):
Looks good, a few questions.

neutron/neutron.go
hmlanigan: PortSecurityEnabled is a *bool because it's optional depending on whether you are using the ML2 extension to neutron.  If it's nil, you don't aren't using the extension.  External is not part of an extension.

nova/nova.go
hmlanigan: yes, done.

testservices/novaservice/service.go
hmlanigan: done.

hmlanigan:
$$merge$$

PR #45, opened 2017-05-01
Add Fault to nova.ServerDetail for details of a nova instance in an E…

hmlanigan:
…RROR state.

Update testservices to allow for 'No valid host' error on specified availability zone.  ServerDetails will now return with server StatusActive, unless StatusError triggered with AZ.

Also tested against a live openstack where one network caused a 'No Valid Host' error.

### DIFF
--- a/testservices/novaservice/service_test.go
+++ b/testservices/novaservice/service_test.go
@@ -241,6 +241,7 @@ func (s *NovaSuite) TestAddServerWithLinks(c *gc.C) {
 	s.createServer(c, server)
 	defer s.deleteServer(c, server)
 	sr, _ := s.service.server(server.Id)
+	server.Status = nova.StatusActive
 	c.Assert(*sr, gc.DeepEquals, server)
 }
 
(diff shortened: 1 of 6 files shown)


testservices/novaservice/service_test.go
```diff
@@ -252,29 +252,6 @@ func (s *NovaSuite) TestAddServerTwiceFails(c *gc.C) {
 	c.Assert(err, gc.ErrorMatches, `conflictingRequest: A server with id "test" already exists`)
 }
 
-// A control point can be used to change the status of the added server.
-func (s *NovaSuite) TestAddServerControlPoint(c *gc.C) {
```
babbageclunk: Why was this deleted? Add the reason in the description if it was a driveby tidyup.

testservices/novaservice/service_test.go
```diff
@@ -417,6 +397,7 @@ func (s *NovaSuite) TestAllServersWithMultipleFilters(c *gc.C) {
 	for _, server := range servers {
 		s.createServer(c, server)
 		defer s.deleteServer(c, server)
+		//server.Status = nova.StatusActive
```
babbageclunk: Deliberate? Worth an explanatory comment if so.

babbageclunk (approved):
Looks good, a couple of questions again.

testservices/novaservice/service_test.go
hmlanigan: nope, removed.

testservices/novaservice/service_test.go
hmlanigan: Updated the test to change the server name via the hook.  Changing the server status will 
no longer work due to changes to the test service to mimic the behavior of server status
of building, active and error.

hmlanigan:
$$merge$$

PR #46, opened 2017-05-03
Fix lp:1682827.  If error 503 and "Retry-After" received, retry request.

hmlanigan:
Add related tests.

### DIFF
--- a/http/client.go
+++ b/http/client.go
@@ -290,7 +290,16 @@ func (c *Client) sendRateLimitedRequest(
 		if err != nil {
 			return nil, errors.Newf(err, "failed executing the request %s", URL)
 		}
-		if resp.StatusCode != http.StatusRequestEntityTooLarge || resp.Header.Get("Retry-After") == "" {
+
+		switch resp.StatusCode {
+		case http.StatusRequestEntityTooLarge,
+			http.StatusForbidden,
+			http.StatusServiceUnavailable,
+			http.StatusTooManyRequests:
+			if resp.Header.Get("Retry-After") == "" {
+				return resp, nil
+			}
+		default:
 			return resp, nil
 		}
 		resp.Body.Close()
--- a/nova/local_test.go
+++ b/nova/local_test.go
@@ -94,12 +94,12 @@ func (s *localLiveSuite) TearDownTest(c *gc.C) {
 
 // Additional tests to be run against the service double only go here.
 
-func (s *localLiveSuite) retryLimitHook(sc hook.ServiceControl) hook.ControlProcessor {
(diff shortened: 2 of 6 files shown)


http/client.go
```diff
@@ -290,7 +290,9 @@ func (c *Client) sendRateLimitedRequest(
 		if err != nil {
 			return nil, errors.Newf(err, "failed executing the request %s", URL)
 		}
-		if resp.StatusCode != http.StatusRequestEntityTooLarge || resp.Header.Get("Retry-After") == "" {
+		if (resp.StatusCode != http.StatusRequestEntityTooLarge &&
+			resp.StatusCode != http.StatusServiceUnavailable) ||
```
axw: according to the bug, and the openstack-compute API docs (https://docs.openstack.org/mitaka/config-reference/compute/api.html), we should be looking for 403 (http.StatusForbidden), not 503 (http.StatusServiceUnavailable)

(While neither is correct, 503 does seem the more sensible of the two. Are the docs wrong?)

wallyworld: The bug report has the error as having code 503. So we need to cater for that.
The docs seem to say that mitaka uses 403, so that enters the mix as well.
Old openstacks use 413, so add that to the list.
And the RFC says 429, so use that too.

IMO, the logic should be:

if statusCode in [403, 413, 429, 503] and header.Retry-After != "" then retey

hmlanigan: Per the bug 1682827, in some cases, openstack is returning an error code 503 with the message "The maximum request receiving rate is exceeded."  So far 503 is undocumented.  

We currently check for 413.  I'll add 403 also.

http/client.go
```diff
@@ -290,7 +292,13 @@ func (c *Client) sendRateLimitedRequest(
 		if err != nil {
 			return nil, errors.Newf(err, "failed executing the request %s", URL)
 		}
-		if resp.StatusCode != http.StatusRequestEntityTooLarge || resp.Header.Get("Retry-After") == "" {
+
+		retryAfterStatusCodes := set.NewInts(
```
axw: this would be a bit more efficient as a switch:

```go
switch resp.StatusCode {
case http.StatusRequestEntityTooLarge,
         http.StatusForbidden,
         http.StatusServiceUnavailable,
         http.StatusTooManyRequests:
    if resp.Header.Get("Retry-After") == "" {
        return resp, nil
    }
default:
    return resp, nil
}
```

axw (approved):

http/client.go
hmlanigan: updated.

hmlanigan:
$$merge$$

PR #47, opened 2017-05-03
Fix bug lp:1683495, if nova ServerDetail.Image fails to unmarshall as

hmlanigan:
a nova.Entity, try as a string before failing.  Added related test.

### DIFF
--- a/nova/json.go
+++ b/nova/json.go
@@ -88,7 +88,16 @@ func (entity *Entity) UnmarshalJSON(b []byte) error {
 	var je jsonEntity = jsonEntity(*entity)
 	var err error
 	if err = json.Unmarshal(b, &je); err != nil {
-		return err
+		// Related Bug lp:1683495, within the openstack compute api
+		// response to List Servers Detailed, the image object might
+		// be an empty string when you boot the server from a volume.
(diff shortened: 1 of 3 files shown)


nova/json.go
```diff
@@ -88,7 +88,17 @@ func (entity *Entity) UnmarshalJSON(b []byte) error {
 	var je jsonEntity = jsonEntity(*entity)
 	var err error
 	if err = json.Unmarshal(b, &je); err != nil {
-		return err
+		// Related Bug lp:1683495, within the openstack compute api
+		// response to List Servers Detailed, the image object might
+		// be an empty string when you boot the server from a volume.
+		// Before failing here, let's check to see if that's the case.
+		err2 := json.Unmarshal(b, &je.Id)
```
axw: Maybe just test `if string(b) == "\"\""`? Feels a little odd to unmarshal into a field when we know it can only be an empty string.

axw (approved):

nova/json.go
hmlanigan: done

hmlanigan:
$$merge$$

hmlanigan:
fixed unrelated test error (bug 49) to pass tests.

$$merge$$

PR #50, opened 2017-05-25
implement a swift object ReadSeeker

jrwren:


jrwren:
I'm curious if something like this would be welcome to goose.

http/client.go
```diff
@@ -212,6 +212,7 @@ func (c *Client) BinaryRequest(method, url, token string, reqData *RequestData,
 	if err != nil {
 		return
 	}
+	reqData.RespValue = resp.StatusCode
```
mhilton: This seems to me to be rather an abuse of the RespValue field, perhaps add a new field to the RespStatus if that is required.

identity/legacy.go
```diff
@@ -23,11 +23,15 @@ func (l *Legacy) Auth(creds *Credentials) (*AuthDetails, error) {
 	request.Header.Set("X-Auth-User", creds.User)
 	request.Header.Set("X-Auth-Key", creds.Secrets)
 	response, err := l.client.Do(request)
-	defer response.Body.Close()
+	defer func() {
```
mhilton: Normally this is done with
```go
if err != nil {
	return nil, err
}
defer response.Body.Close()
```

swift/swift.go
```diff
@@ -109,6 +110,21 @@ func (c *Client) GetReader(containerName, objectName string) (io.ReadCloser, htt
...
 }
 
+// GetReadHandle creates a handler to the specified object's data.
+func (c *Client) GetReadHandle(containerName, objectName string) (io.ReadSeeker, http.Header, error) {
+	requestData := goosehttp.RequestData{}
+	err := c.touchObject(&requestData, client.HEAD, containerName, objectName)
+	if err != nil {
+		return nil, nil, err
+	}
+	l, err := strconv.ParseInt(requestData.RespHeaders.Get("Content-Length"), 10, 64)
```
mhilton: It seems a shame that this has to be done as the content length will have already been parsed  once.

swift/swift.go
```diff
@@ -188,3 +204,67 @@ func maybeNotFound(err error, format string, arg ...interface{}) error {
...
+		return
+	}
+	defer requestData.RespReader.Close()
+	readSize := 0
+	for n < len(p) {
+		readSize, err = requestData.RespReader.Read(p[n:])
+		n += readSize
+		o.pos += int64(readSize)
+		if err != nil {
+			if err == io.EOF || err == io.ErrUnexpectedEOF {
```
mhilton: This seems odd, whilst it's reasonable to treat EOF as not an error you are worried about, ErrUnexpectedEOF should probably never be ignored like this. If it is really the behavior you want then perhaps add a comment.

swift/swift.go
```diff
@@ -188,3 +204,67 @@ func maybeNotFound(err error, format string, arg ...interface{}) error {
...
+				o.pos = o.length
+			}
+			return
+		}
+	}
+	return
+}
+
+// Seek sets the offset for the next Read operation on the handler.
+func (o *ohandler) Seek(offset int64, whence int) (int64, error) {
```
mhilton: The contract for seeker suggests that

> Seeking to an offset before the start of the file is an error. Seeking to any positive offset is legal, but the behavior of subsequent I/O operations on the underlying object is implementation-dependent.

I think implementing that behavior would be worthwhile.

testservices/swiftservice/service_http.go
```diff
@@ -215,3 +222,84 @@ func (s *Swift) SetupHTTP(mux *http.ServeMux) {
 func (s *Swift) Stop() {
 	// noop
 }
+
+// copied from net/http
```
mhilton: This should probably be in it's own file then with the original license header.

mhilton (changes requested):
I'm worried that a few things here are abusing existing values rather than adding new ones as needed.

identity/legacy.go
jrwren: I'll try. I ran into cases where this would cause panic because something was nil and dereferenced.

swift/swift.go
jrwren: I agree. I could copy ContentLength from net/http.Response into a new field in goose client. 

I do not understand entire existence of goose client hiding the underlying net/http.Response. It seems requestData's Resp* fields should instead be the net/http.Response. Since I cannot understand its design, I don't know the right fix here. Please someone tell me what to do.

swift/swift.go
jrwren: Done.

http/client.go
jrwren: Oops, debug code leaked in. I removed this change.

fabricematrat (commented):
LGTM

http/client.go
```diff
@@ -72,15 +72,16 @@ func unmarshallError(jsonBytes []byte) (*ErrorResponse, error) {
...
-	RespReader     io.ReadCloser
-	RespHeaders    http.Header
+	ReqHeaders        http.Header
+	Params            *url.Values
+	ExpectedStatus    []int
+	ReqValue          interface{}
+	RespValue         interface{}
+	ReqReader         io.Reader
+	ReqLength         int
+	RespContentLength int64
```
mhilton: Let's make this RespLength to match the field above.

swift/swift.go
```diff
@@ -188,3 +200,72 @@ func maybeNotFound(err error, format string, arg ...interface{}) error {
...
+
+type oreadseeker struct {
+	c             *Client
+	containerName string
+	objectName    string
+	length        int64
+	pos           int64
+}
+
+// Read reads up to len(b) bytes from the oreadseeker. It returns the number of bytes read and any error encountered. At end of stream, Read returns no EOF on the first EOF and EOF after that.
```
mhilton: please put some newlines in this comment.

mhilton (approved):
LGTM with a couple of minor thoughts.

jrwren:
$$merge$$

mhilton:
$$merge$$

PR #51, opened 2017-07-04
swift: improve GetReadSeeker

rogpeppe:
We change the return value so that it returns
a ReadSeekCloser, opening the way for
having state inside the returned reader.

We also change it to ensure that the read
fails if the content changes underfoot,
so we can rely on the length remaining the same.

We also tidy up the live tests a little and fix a failure
(they now pass), and change the Swift test server
so that it deals with etags and If-Match headers
correctly by using http.ServeContent.


mhilton (approved):
LGTM

fabricematrat (approved):

swift/live_test.go
```diff
@@ -294,20 +300,41 @@ func (s *LiveTests) TestReadSeekerReadLimit(c *gc.C) {
...
 	_, err = r.Seek(-1, io.SeekStart)
 	c.Check(err, gc.NotNil)
 	_, err = r.Seek(-20, io.SeekEnd)
 	c.Check(err, gc.NotNil)
 }
+
+func (s *LiveTests) TestReadSeekerFileChangedUnderfoot(c *gc.C) {
+	object := "test_obj2"
+	data := "...some data..."
+	err := s.swift.PutReader(s.containerName, object, bytes.NewReader([]byte(data)), int64(len(data)))
```
jameinel: Why PutReader vs PutObject? The latter seems a lot easier to write.

swift/swift.go
```diff
@@ -111,14 +116,19 @@ func (c *Client) GetReader(containerName, objectName string) (io.ReadCloser, htt
 }
 
 // GetReadSeeker creates a read seeker closer to the specified object's data.
-func (c *Client) GetReadSeeker(containerName, objectName string) (io.ReadSeeker, http.Header, error) {
+func (c *Client) GetReadSeeker(containerName, objectName string) (ReadSeekCloser, http.Header, error) {
```
jameinel, line 119: Given the name of the type was in the method name (GetReadSeeker), it feels a bit odd to not have them match. That said, "GetReadSeeker" was not a good name to start with.
Just 'GetReader' comes to mind to indicate it is a type you can pull data out of, rather than concrete bytes (eg, GetReader vs GetBytes or GetString)

swift/swift.go
```diff
@@ -111,14 +116,19 @@ func (c *Client) GetReader(containerName, objectName string) (io.ReadCloser, htt
...
 	requestData := goosehttp.RequestData{}
 	err := c.touchObject(&requestData, client.HEAD, containerName, objectName)
 	if err != nil {
 		return nil, nil, err
 	}
-	o := &oreadseeker{c, containerName, objectName, requestData.RespLength, 0}
-	return o, requestData.RespHeaders, err
+	return &readSeeker{
```
jameinel: Having a 'readSeeker' actually be a ReadSeekCloser is also likely to add to confusion.
swiftFile perhaps?

swift/swift.go
```diff
@@ -201,56 +210,63 @@ func maybeNotFound(err error, format string, arg ...interface{}) error {
...
-	if o.pos == o.length {
+func (o *readSeeker) Read(p []byte) (n int, err error) {
+	if o.pos >= o.length {
 		return 0, io.EOF
 	}
-	end := int64(len(p)) + o.pos
+	end := o.pos + int64(len(p))
+	if end > o.length {
+		p = p[:o.length-o.pos]
+		end = o.pos + int64(len(p))
```
jameinel: if len(p) is defined as o.length-o.pos
then o.pos + o.length - o.pos => o.length
Can you just do:
end = o.length
and have it be clearer?

swift/swift.go
```diff
@@ -201,56 +210,63 @@ func maybeNotFound(err error, format string, arg ...interface{}) error {
...
+		header.Set("If-Match", o.etag)
+	}
 	requestData := goosehttp.RequestData{
+		// Signal to the client that we want the reader back
+		// by placing a dummy reader there.
+		// (RespReader gets replaced).
 		RespReader:     &emptyReadCloser,
 		ReqHeaders:     header,
-		ExpectedStatus: []int{http.StatusOK, http.StatusPartialContent},
+		ExpectedStatus: acceptableReadStatuses,
```
jameinel: Calling the value "acceptableReadStatuses" feels a bit off. Something like "supportedReadStatuses" might be clearer that we know we can handle X and Y.
I don't fully understand why it gets pulled into a package variable, given it seems very tied to exactly what this function can actually handle.

swift/swift.go
```diff
@@ -201,56 +210,63 @@ func maybeNotFound(err error, format string, arg ...interface{}) error {
...
-			return
-		}
+	if requestData.RespStatusCode == http.StatusPreconditionFailed {
+		return 0, goerrors.New("file has changed since it was opened")
 	}
-	return
+	// Since we should be ensuring that the data never changes
+	// underfoot because we're setting the If-Match header,
+	// then the length should always be the same as when we
+	// first opened it, so any short read justifies an ErrUnexpectedEOF.
```
jameinel, line 262: You're not unconditionally setting If-Match, you're only doing so *if* you have an Etag. Are you sure you'll always have an etag?

testservices/swiftservice/service_http.go
```diff
@@ -135,14 +139,18 @@ func (s *Swift) handleObjects(container, object string, w http.ResponseWriter, r
...
 	case "GET":
-		w.Header().Set("Content-Type", "application/json; charset=UF-8")
-		w.Header().Set("Content-Length", strconv.Itoa(len(objdata)))
-		w.WriteHeader(http.StatusOK)
-		if hr, err := parseRange(r.Header.Get("Range"), int64(len(objdata))); err == nil && len(hr) > 0 {
-			w.Write([]byte(objdata[hr[0].start : hr[0].start+hr[0].length]))
-		} else {
-			w.Write([]byte(objdata))
-		}
+		// Note: even though the real Swift service does not
```
jameinel, line 142: You're saying that Openstack Swift doesn't quote its ETAG? weird.

jameinel (commented):
some naming things aside (readSeeker actually being a readSeekCloser, etc.)
I think the general changes are fine.
I'm a little concerned that just dropping the ErrUnexpectedEOF may work for a particular client, but not for all.

testservices/swiftservice/service_http.go
rogpeppe: Yup, it's weird (and wrong).

swift/swift.go
rogpeppe: We should always have an Etag, yes, but I think an ErrUnexpectedEOF in the unlikely event that we don't and the content changes underfoot seems OK to me.

swift/swift.go
rogpeppe: good point. done.

swift/swift.go
rogpeppe: I pulled it into a package variable to avoid allocating a new slice every time. 
I don't see too much difference in the names, so *shrug* done.

swift/swift.go
rogpeppe: sure. done-ish (named it "object" as swift has objects, not files, and the "swift" seems redundant given that we're in the swift package).

swift/swift.go
rogpeppe: In a subsequent PR, I'm planning to remove GetReadSeeker (it's new and only used in one place) and add OpenObject that returns a ReadSeekCloser, and make GetReader use that.
Alternatively, I could change the signature of GetReader but I'd prefer not to break backward compatibility for a method that's been around for a long time.

(there is already a GetReader method)

swift/live_test.go
rogpeppe: done

rogpeppe: done

jameinel (approved):

rogpeppe:
$$merge$$

mhilton:
$$merge$$

PR #52, opened 2017-07-05
http: persistent client connections

rogpeppe:
Two issues were preventing client connections from being
reused (leading to significant performance reduction):

- we were specifically asking for connections not to be reused,
citing Go issue 4677, which has been fixed since Go 1.6

- Some responses (notably the response to CreateContainer)
were returning response bodies which were not being read.

### DIFF
--- a/http/client.go
+++ b/http/client.go
@@ -26,17 +26,6 @@ const (
 	contentTypeOctetStream = "application/octet-stream"
 )
 
-func init() {
-	// See https://code.google.com/p/go/issues/detail?id=4677
-	// We need to force the connection to close each time so that we don't
-	// hit the above Go bug.
-	roundTripper := http.DefaultClient.Transport
-	if transport, ok := roundTripper.(*http.Transport); ok {
-		transport.DisableKeepAlives = true
-	}
-	http.DefaultTransport.(*http.Transport).DisableKeepAlives = true
-}
-
 type Client struct {
 	http.Client
 	maxSendAttempts int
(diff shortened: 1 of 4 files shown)


frankban (approved):
Looks good. Please add license headers to the two new files.

http/client.go
```diff
@@ -218,6 +207,16 @@ func (c *Client) BinaryRequest(method, url, token string, reqData *RequestData,
...
 	} else {
+		if method != "HEAD" && resp.ContentLength != 0 {
+			// Read a small amount of data from the response
+			// body so that the client connection can
+			// be reused.
+			size := resp.ContentLength
+			if size > 1024 || size < 0 {
+				size = 1024
+			}
+			resp.Body.Read(make([]byte, size))
```
jameinel, line 218: Don't you have to read all of ContentLength for the connection to be ready for the next response?

http/conn_reuse_test.go
```diff
@@ -0,0 +1,73 @@
...
+		c.Assert(req.RespHeaders.Get("Test"), gc.Equals, expectTestHeader)
+	}
+	for i, path := range []string{
+		"/with-body-chunked",
+		"/with-body-non-chunked",
+		"/no-body",
+	} {
+		assertReq(path, fmt.Sprint(i))
+	}
+	c.Assert(connCount, gc.Equals, int32(1))
```
jameinel, line 67: if you're going to use atomic to AddInt, don't you have to use LoadInt to be sure this thread gets the right content?

jameinel (approved):
Assuming you've tested this live, since I'm sure that's how we encountered issue 4677 originally.

http/conn_reuse_test.go
rogpeppe: I'm assuming that connections won't be made without an HTTP request to stimulate them. The atomic.Add is probably overkill, tbh, as I can't see net/http making two concurrent connections with only one request. If there really is a danger of the connection count incrementing asynchronously after all the connections are made, I think I'd prefer the race detector to complain.

http/client.go
rogpeppe: Yes, but I don't think it's worth reading large response bodies to save a connection. In the particular case we're seeing in swift, the body is only about 100 bytes. We could tweak this value later if we find it's not effective.

rogpeppe:
$$merge$$

mhilton:
$$merge$$

PR #53, opened 2017-07-06
swift: readahead support when reading objects

rogpeppe:
We make it more efficient for streaming and seeking
readers by issuing reads ahead of time, meaning we
can be processing some data while the HTTP request
is being read.

Control over the amount of readahead is governed by
the readAhead parameter to the new OpenObject
method, which subsumes GetReadSeeker.

### DIFF
--- a/internal/httpfile/httpfile.go
+++ b/internal/httpfile/httpfile.go
@@ -1,0 +1,531 @@
+package httpfile
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"strconv"
+	"strings"
+)
+
+// NoSkip holds the maximum number of bytes that we
+// will read instead of discarding a connection in order
+// to seek.
+const NoSkip = 64 * 1024
+
+// unlimited is used as a special case to signify an unlimited number of
+// bytes. We trust that no files are as big as 4 exabytes for now.
+const unlimited = int64(1 << 62)
+
+// Request represents a request made by a File instance.
+// It will usually be mapped into an HTTP request
+// to the object that has been opened.
+type Request struct {
+	Method string
+	Header http.Header
+}
+
+// Response represents a response from a Request.
+// It will usually be created from an HTTP response.
+type Response struct {
+	StatusCode    int
+	Header        http.Header
+	ContentLength int64
+	Body          io.ReadCloser
+}
+
+type Client interface {
+	// Do sends an HTTP request to the file.
+	// The provided header fields are added
+	// to the request, and the given request is
+	// send with the given method (GET or HEAD).
+	Do(req *Request) (*Response, error)
+}
+
+// SupportedResponseStatuses holds the set of HTTP statuses that a
+// response may legimately have.
+var SupportedResponseStatuses = []int{
+	http.StatusOK,
+	http.StatusNotFound,
+	http.StatusPartialContent,
+	http.StatusPreconditionFailed,
+}
+
+var (
+	// ErrNotFound is returned by Open when the file
+	// does not exist.
+	ErrNotFound = errors.New("file not found")
+
+	// errChanged is returned by a readerAhead
+	errChanged = errors.New("file has changed since it was opened")
+
+	// errOutOfRange is returned by readerAhead to signify
+	// that a given read request is out of the range of
+	// available data.
+	errOutOfRange = errors.New("read out of range")
+)
+
+type File struct {
+	client Client
+
+	// length holds the length of the file.
+	// While initializing the file, this is -1
+	// to signify that the length is unknown.
+	length int64
+
+	// etag holds the entity tag of the file, as returned by the
+	// server in the Etag header. Note that this includes the double
+	// quotes around it too.
+	etag string
+
+	// readAhead holds the number of bytes
+	// to request in advance of any Read.
+	readAhead int64
+
+	// pos holds the current seek position of the reader.
+	pos int64
+
+	// reader0 represents the currently running GET request.
+	reader0 *readerAhead
+
+	// reader1 represents the speculative GET request
+	// one block ahead of reader0. This is never
+	// non-nil when reader0 is nil.
+	reader1 *readerAhead
+}
+
+// Open opens a new file that uses the given client to issue read
+// requests. It returns the open file and any HTTP header returned by
+// the initial client request.
+//
+// If the file is not found, it returned ErrNotFound.
+//
+// The readAhead parameter governs the amount of data that will be
+// requested before any Read calls are made. If it's zero, no data will
+// be requested before Read calls are made. If it's -1, unlimited data
+// will be requested; otherwise the given number of bytes will be
+// requested.
+//
+// If readAhead is less than NoSkip and not -1, all HTTP connections
+// will be available for reuse regardless of how Seek is used.
+func Open(c Client, readAhead int64) (*File, http.Header, error) {
+	if readAhead == -1 {
+		readAhead = unlimited
+	}
+	f := &File{
+		client:    c,
+		length:    -1, // Unknown.
+		readAhead: readAhead,
+	}
+	ra := f.newReaderAhead(0, readAhead)
+	if err := ra.wait(); err != nil {
+		if err == errChanged {
+			// OpenStack can return StatusPreconditionFailed
+			// when the container or object does not exist,
+			// so return ErrNotFound in that case.
+			return nil, ra.header, ErrNotFound
+		}
+		return nil, ra.header, err
+	}
+	f.reader0 = ra
+	f.etag = ra.header.Get("Etag")
+	f.length = ra.length
+	return f, ra.header, nil
+}
+
+// Size returns the total size of the file.
+func (f *File) Size() int64 {
+	return f.length
+}
+
+// Read implements io.Reader.Read.
+func (f *File) Read(buf []byte) (int, error) {
+	if f.pos >= f.length {
+		return 0, io.EOF
+	}
+	if f.reader0 == nil {
+		f.reader0 = f.newReaderAhead(f.pos, f.pos+int64(len(buf)))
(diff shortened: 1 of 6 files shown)


internal/httpfile/httpfile.go
```diff
@@ -0,0 +1,429 @@
...
+	// only valid to read after the done channel
+	// has been closed.
+	r          io.ReadCloser
+	header     http.Header
+	length     int64
+	statusCode int
+	err        error
+}
+
+const infinity = int64(1 << 62)
```
jrwren: naming quip: I prefer MaxInt64 or VeryLargeInt to infinity here. Its confusing to read because I think it is referring to  float inf.

internal/httpfile/httpfile.go
```diff
@@ -0,0 +1,429 @@
...
+		header.Set("If-Match", f.etag)
+	}
+	return req
+}
+
+// checkContentRange checks that the content range in the response is as
+// expected. If a server doesn't support range requests for whatever
+// reason, this should catch it.
+//
+// It returns the length of the entire
```
jrwren: ... cliffhanger. I'm on the edge of my seat. entire...

jrwren (approved):
LGTM as long as it works and performs as good or better than previous.

jrwren:
My quick tests show it as performing much better, but this is using the behavior of io.Copy and io.Discard which has a certain pattern. Other patterns may vary.

Old method:
reading a small  1193B file, 1000 times takes: 19.111059 seconds
reading a larger 111062826B file, 1 time takes: 204.638050 seconds


```
	start := time.Now()
	n := int64(0)
	for i := 0; i < *count; i++ {
		req2, _, err := s.OpenObject(*container, *name, -1)
		if err != nil {
			fmt.Printf("ERROR:%s\n", err)
			return
		}
		m, err := io.Copy(ioutil.Discard, req2)
		if err != nil {
			fmt.Printf("ERROR:%s\n", err)
			return
		}
		n += m
		req2.Close()
	}
	fmt.Printf("%d bytes read in %f seconds\n", n, time.Now().Sub(start).Seconds())
```

New Method:

Reading a small 1193B file, 1000 times takes: 11.413s
1193000 bytes read in 11.413108 seconds

reading a larger 111062826B file, 1 time takes: 12.6s
111062826 bytes read in 12.678619 seconds

rogpeppe:
Thanks for the measurements? What value for readAhead were you using? It could make a big difference (and readAhead=-1) is essentially equivalent to the old GetReader except you can still seek if you want to.

internal/httpfile/httpfile.go
```diff
@@ -0,0 +1,429 @@
...
+	// one block ahead of reader0. This is never
+	// non-nil when reader0 is nil.
+	reader1 *readerAhead
+}
+
+var supportedReadStatuses = []int{http.StatusOK, http.StatusPartialContent, http.StatusPreconditionFailed}
+
+var errOutOfRange = errors.New("read out of range")
+
+func Open(c Client, readAhead int64) (*File, http.Header, error) {
```
mhilton: doc comment

mhilton: also why are you returning headers here?

internal/httpfile/httpfile.go
mhilton: this really needs a comment.

internal/httpfile/httpfile.go
```diff
@@ -0,0 +1,429 @@
...
+func (ra *readerAhead) close() {
+	go func() {
+		<-ra.done
+		if ra.r != nil {
+			ra.r.Close()
+		}
+	}()
+}
+
+func (f *File) newRequest(p0, p1 int64) *Request {
```
mhilton: This could do with a doc comment, especially with the behavior of p1 which is non-obvious.

internal/httpfile/httpfile.go
mhilton: entire what?

internal/httpfile/httpfile.go
```diff
@@ -0,0 +1,429 @@
...
+			return 0, fmt.Errorf("received unexpected Content-Range %q in response", got)
+		}
+		if resp.ContentLength == -1 {
+			return 0, fmt.Errorf("unknown file length in response")
+		}
+		return resp.ContentLength, nil
+	}
+	r, ok := parseContentRange(got)
+	if !ok {
+		return 0, fmt.Errorf("bad Content-Length header %q", got)
```
mhilton: Content-Range?

internal/httpfile/httpfile.go
```diff
@@ -0,0 +1,429 @@
...
+	}
+	s, ok = trimPrefix(s, "-")
+	if !ok {
+		return contentRange{}, false
+	}
+	r.p1, s, ok = parseInt(s)
+	if !ok {
+		return contentRange{}, false
+	}
+	// Use the usual Go convention for half-open ranges.
```
mhilton, line 491: This should not be buried this far down. please put a comment on contentRange making it clear what p1 is.

internal/httpfile/httpfile.go
```diff
@@ -0,0 +1,429 @@
...
+	r.p1++
+	s, ok = trimPrefix(s, "/")
+	if !ok {
+		return contentRange{}, false
+	}
+	r.length, s, ok = parseInt(s)
```
mhilton, line 497: The length can legitimately be "*" under some circumstances, although probably not any we'll see.

internal/httpfile/httpfile.go
```diff
@@ -0,0 +1,429 @@
...
+func Open(c Client, readAhead int64) (*File, http.Header, error) {
+	if readAhead == -1 {
+		readAhead = infinity
+	}
+	f := &File{
+		client:    c,
+		length:    -1, // Unknown.
+		readAhead: readAhead,
+	}
+	f.reader0 = f.newReaderAhead(0, readAhead)
```
mhilton: I suspect that if you always did a HEAD request here to get the length and Etag etc. then a number of the special cases in the code below would disappear making things simpler.

swift/swift.go
```diff
@@ -197,6 +178,49 @@ func (c *Client) SignedURL(containerName, file string, expires time.Time) (strin
...
+		object:    objectName,
+	}
+	f, h, err := httpfile.Open(client, readAhead)
+	if err != nil {
+		return nil, h, err
+	}
+	return f, h, nil
+}
+
+// GetObject retrieves the specified object's data.
```
mhilton: GetReader?

swift/swift.go
```diff
@@ -210,80 +234,40 @@ func maybeNotFound(err error, format string, arg ...interface{}) error {
...
+		ContentLength: gooseReq.RespLength,
+		Body:          gooseReq.RespReader,
 	}
-	o.pos = offset
-	return o.pos, nil
-}
-
-func (o *object) Close() error {
-	return nil
+	//log.Printf("%#v -> %#v", req, resp)
```
mhilton: I don't understand this comment.

mhilton (approved):
This is fine, with a few changes. However I think it's probably more complex than it needs to be. I'd be interested how much less overhead this has than just dropping the connection on every seek. Also it might be worth just buffering the whole object when its sufficiently small so that seeking has very little cost.

internal/httpfile/httpfile.go
rogpeppe: doc comment done. the headers are returned because the goose client returns the headers from its GetReader method for some odd reason.

internal/httpfile/httpfile.go
rogpeppe: You're right that it would make the code simpler, but I think it's worth pursuing the goal of issuing a single request when reasonable.

internal/httpfile/httpfile.go
rogpeppe: renamed to "unlimited" and added a comment.

internal/httpfile/httpfile.go
rogpeppe: done

internal/httpfile/httpfile.go
rogpeppe: I refactored this function into contentRangeFromResponse, which seems to
make the semantics clearer. The comment has gone.

internal/httpfile/httpfile.go
rogpeppe: done

internal/httpfile/httpfile.go
rogpeppe: done

internal/httpfile/httpfile.go
rogpeppe: added a comment

swift/swift.go
rogpeppe: done

swift/swift.go
rogpeppe: removed

internal/httpfile/httpfile.go
```diff
@@ -283,19 +355,45 @@ func (ra *readerAhead) readAt(buf []byte, p0 int64) (int, error) {
...
+	ra.p0 += n
+	if err == io.EOF {
+		return io.ErrUnexpectedEOF
+	}
+	return err
+}
+
+// newRequest returns a new Request to read the data
+// in the interval [p0, p1). If p1 is unlimited, an
+// unlimited of data is requested.
```
mhilton: unlimited amount ?

mhilton (approved):
LGTM with a small nitpick

rogpeppe:
$$merge$$

mhilton:
$$merge$$

hmlanigan (commented):
goose.v2 is currently listed as experimental.  I was in the process of making compatibility breaking changes... which I'd like to get back to soonish.  Given that, there is no need to keep methods like GetRead() if we don't want.

LGTM

PR #55, opened 2017-08-23
avoid arbitrary memory use in upload

rogpeppe:
When uploading a file, the code was reading the entire
body into memory before making the request.
This caused out-of-memory errors when
uploading large files.

Now, if you provide a ReadSeeker, it will seek to
the start of the file rather than reading the whole
thing into memory.

It's tempting to change the API so that it requires
a ReadSeeker rather than a Reader, but that would break
the goose.v2 compatibility guarantee.

### DIFF
--- a/http/client.go
+++ b/http/client.go
@@ -24,6 +24,10 @@ import (
 const (
 	contentTypeJSON        = "application/json"
 	contentTypeOctetStream = "application/octet-stream"
+	// maxBufSize holds the maximum amount of data
+	// that may be allocated as a buffer when sending
+	// a request when the body is not seekable.
+	maxBufSize = 1024 * 1024 * 1024
 )
 
 type Client struct {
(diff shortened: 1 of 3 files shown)


http/client.go
```diff
@@ -272,28 +276,52 @@ func (c *Client) sendRequest(
...
+			}{notifier, reqReader}
 		}
-		req, err := http.NewRequest(method, URL, reqReader)
+		req, err := http.NewRequest(method, URL, nil)
 		if err != nil {
-			err = errors.Newf(err, "failed creating the request %s", URL)
-			return nil, err
+			return nil, errors.Newf(err, "failed creating the request %s", URL)
 		}
+		req.Body = body
```
axw: please add a comment stating why you're setting req.Body, rather than passing it into NewRequest

axw (approved):

http/client.go
rogpeppe: Good catch. It turns out that it wasn't necessary in fact, as NewRequest does a dynamic type check for io.ReadCloser.

I've also added some more tests that check the new logic - there were some significant gaps in coverage previously.

rogpeppe:
$$merge$$

mhilton:
$$merge$$

PR #57, opened 2017-10-27
To allow for fix of lp:1707248 - Adding filter to ListFloatingIPsV2()

hmlanigan:
Add filter to ListFloatingIPsV2 and neutron.FilterProjectId, so we can get a list of FloatingIPs within the client's project.  Neutron testing and test service updated accordingly.

### DIFF
--- a/neutron/neutron.go
+++ b/neutron/neutron.go
@@ -26,6 +26,7 @@ const (
 const (
 	FilterRouterExternal = "router:external" // The router:external
 	FilterNetwork        = "name"            // The network name.
+	FilterProjectId      = "project_id"      // The project id
 )
 
 // NetworkV2 contains details about a labeled network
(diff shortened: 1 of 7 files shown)


neutron/neutron.go
```diff
@@ -160,12 +161,15 @@ type FloatingIPV2 struct {
...
-func (c *Client) ListFloatingIPsV2() ([]FloatingIPV2, error) {
+func (c *Client) ListFloatingIPsV2(filter ...*Filter) ([]FloatingIPV2, error) {
 	var resp struct {
 		FloatingIPV2s []FloatingIPV2 `json:"floatingips"`
 	}
-
-	requestData := goosehttp.RequestData{RespValue: &resp}
+	var params *url.Values
+	if len(filter) > 0 {
+		params = &filter[0].v
```
wallyworld, line 175: If we're only ever going to use just one Filter value we should return an error if len(filter) > 1 (and add a test)
Otherwise it's confusing reading the code.
The doc comment also should be updated to mention the filter and expected semantics.

The other option:

ListFloatingIPsV2(filter *Filter) ([]FloatingIPV2, error)

if filter != nil {
  params = &filter.v
}

But this will change the caller to make it pass in nil when there's no filter. We have done that in other places, so whatever works best....

wallyworld (approved):
A small change but looks ok otherwise.

neutron/neutron.go
hmlanigan: The filter for ListFloatingIPsV2() is working the same as for ListNetworksV2().  I did update the comments.

hmlanigan:
$$merge$$

PR #58, opened 2017-10-31
When handling ratelimit in http, Retry-After can be in the http-date

hmlanigan:
format as well as delay-seconds.  Handle accordingly.

### DIFF
--- a/http/client_test.go
+++ b/http/client_test.go
@@ -175,7 +175,9 @@ func (s *HTTPClientTestSuite) TestJSONRequestSetsToken(c *gc.C) {
 	c.Check(agent, gc.Equals, "token")
 }
 
-func (s *HTTPClientTestSuite) TestRetry(c *gc.C) {
+func (s *HTTPClientTestSuite) testRetryAfter(c *gc.C,
+	retryAfter func(*time.Time, http.ResponseWriter),
+	verifyWait func(time.Time) (time.Duration, bool)) {
 	count := 0
 	var t0 time.Time
 	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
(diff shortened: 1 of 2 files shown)


http/client_test.go
```diff
@@ -244,6 +244,78 @@ func (s *HTTPClientTestSuite) TestResourceLimitExceeded(c *gc.C) {
...
+	count := 0
+	var t0 time.Time
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		data, _ := ioutil.ReadAll(req.Body)
+		c.Check(string(data), gc.Equals, "request body")
+		count++
+		switch count {
+		case 1:
+			t0 = time.Now()
+			t0 = t0.Add(time.Duration(10) * time.Second)
```
wallyworld: Since this test suit is not using test clocks, adding 10s here will make the test suite take a while. Can we make it 1s like in the other test.

http/client_test.go
```diff
@@ -244,6 +244,78 @@ func (s *HTTPClientTestSuite) TestResourceLimitExceeded(c *gc.C) {
 	c.Assert(err, gc.ErrorMatches, `Resource limit exeeded at URL http://.*`)
 }
 
+func (s *HTTPClientTestSuite) TestRetryHttpDate(c *gc.C) {
```
wallyworld: Can't we combine this test with the one above into common code and just pass in a string value to use for the retry after header value.

wallyworld (changes requested):
Code looks good but I think the tests can be improved.

http/client_test.go
hmlanigan: done

http/client_test.go
hmlanigan: done

wallyworld (approved):
Thanks for the test changes!

wallyworld:
$$merge$$

PR #60, opened 2017-12-12
cinder: use goose/http.Client

axw:
Use goose/http.Client in a few methods
of the Cinder client, so we can use the
common goose/errors functions. Not all
methods have been updated; only the ones
necessary to delete and detach volumes.

Also update the Nova code for dealing
with volume attachments to not regard
404 as "not implemented".

### DIFF
--- a/nova/nova.go
+++ b/nova/nova.go
@@ -798,11 +798,6 @@ func (c *Client) AttachVolume(serverId, volumeId, device string) (*VolumeAttachm
 	}
 	url := fmt.Sprintf("%s/%s/%s", apiServers, serverId, apiVolumeAttachments)
 	err := c.client.SendRequest(client.POST, "compute", "v2", url, &requestData)
-	if errors.IsNotFound(err) {
-		return nil, errors.NewNotImplementedf(
-			err, nil, "the server does not support attaching volumes",
-		)
-	}
 	if err != nil {
 		return nil, errors.Newf(err, "failed to attach volume")
 	}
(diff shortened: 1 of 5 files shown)


nova/nova.go
```diff
@@ -798,11 +798,6 @@ func (c *Client) AttachVolume(serverId, volumeId, device string) (*VolumeAttachm
 	}
 	url := fmt.Sprintf("%s/%s/%s", apiServers, serverId, apiVolumeAttachments)
 	err := c.client.SendRequest(client.POST, "compute", "v2", url, &requestData)
-	if errors.IsNotFound(err) {
-		return nil, errors.NewNotImplementedf(
-			err, nil, "the server does not support attaching volumes",
```
jameinel, line 803: What is the way that we could detect if the server just doesn't support volume attachment at all?
Is there a different error we should be handling?

jameinel (approved):
Seems like a reasonable cleanup. I'm a bit curious how we should be conveying information to the user if the underlying cloud didn't actually support the operation. But maybe if a user is trying to use it, then they should know better.

axw:
> Seems like a reasonable cleanup. I'm a bit curious how we should be conveying information to the user if the underlying cloud didn't actually support the operation. But maybe if a user is trying to use it, then they should know better.

Yeah, that was my rationale for removing it: if you're calling the method, you should already know whether or not volumes and volume attachments are available. You can query for API extensions.

I've not come across any documentation specifying a response that will be sent when accessing a missing extension. From Juju's perspective, the worst that could happen is that we'll interpret a missing extension as the volume not being attached. If that's the case, then it couldn't have been attached in the first case.

nova/nova.go
axw: > Is there a different error we should be handling?

Not that I can see. Just query the extensions API endpoint first if needs be.

axw:
$$merge$$

axw:
Build failed: builder ran out of disk space

axw:
$$merge$$

PR #61, opened 2018-02-04
Minimal version of Neutron GBP API

wupeka:
A minimal Neutron GBP API client required to create a Policy Target inside a Policy Target Group.

### DIFF
--- a/neutron/gbp.go
+++ b/neutron/gbp.go
@@ -1,0 +1,108 @@
+// goose/neutron/gbp - Go package to interact with GBP plugin of the
+// OpenStack Network Service (Neutron) API.
+// It currently supports small subset of the API needed to create a
+// policy target.
+// There is no documentation for GBP API so this code is reverse engineered
+// from Python gbp client code.
+
+package neutron
+
+import (
+	"fmt"
+	"net/http"
+	"net/url"
+
+	"gopkg.in/goose.v2/client"
+	"gopkg.in/goose.v2/errors"
+	goosehttp "gopkg.in/goose.v2/http"
+)
+
+const (
+	ApiApplicationPolicyGroupsV2 = "grouppolicy/application_policy_groups"
+	ApiExternalPolicysV2         = "grouppolicy/external_policys"
+	ApiExternalSegmentsV2        = "grouppolicy/external_segments"
+	ApiL2PolicysV2               = "grouppolicy/l2_policys"
+	ApiL3PolicysV2               = "grouppolicy/l3_policys"
+	ApiNatPoolsV2                = "grouppolicy/nat_pools"
+	ApiNetworkServicePolicysV2   = "grouppolicy/network_service_policys"
+	ApiPolicyActionsV2           = "grouppolicy/policy_actions"
+	ApiPolicyClassifiersV2       = "grouppolicy/policy_classifiers"
+	ApiPolicyRulesV2             = "grouppolicy/policy_rules"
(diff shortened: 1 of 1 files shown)


neutron/gbp.go
```diff
@@ -0,0 +1,96 @@
...
+type PolicyTargetV2 struct {
+	Id                  string   `json:"id,omitempty"`
+	Name                string   `json:"name,omitempty"`
+	Description         string   `json:"description,omitempty"`
+	PolicyTargetGroupId string   `json:"policy_target_group_id"`
+	PortId              string   `json:"port_id,omitempty"`
+	FixedIps            []string `json:"fixed_ips,omitempty"`
+}
+
+func (c *Client) ListPolicyTargetsV2(filter ...*Filter) ([]PolicyTargetV2, error) {
```
hmlanigan: Needs comment.  

What keys are valid for the filter?  neutron.go has Filter*Key variables.

neutron/gbp.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	requestData := goosehttp.RequestData{RespValue: &resp, Params: params}
+	err := c.client.SendRequest(client.GET, "network", "v2.0", ApiPolicyTargetsV2, &requestData)
+	if err != nil {
+		return nil, errors.Newf(err, "failed to get list of networks")
+	}
+	return resp.PolicyTargets, nil
+}
+
+func (c *Client) GetPolicyTargetV2(ptId string) (*PolicyTargetV2, error) {
```
hmlanigan: Needs comment.

Is it possible for the string to have a space?

neutron/gbp.go
```diff
@@ -0,0 +1,96 @@
...
+	url := fmt.Sprintf("%s/%s", ApiPolicyTargetsV2, ptId)
+	requestData := goosehttp.RequestData{RespValue: &resp}
+	err := c.client.SendRequest(client.GET, "network", "v2.0", url, &requestData)
+	if err != nil {
+		return nil, errors.Newf(err, "failed to get policy_target_group detail")
+	}
+	return &resp.PolicyTarget, nil
+}
+
+func (c *Client) CreatePolicyTargetV2(pt PolicyTargetV2) (*PolicyTargetV2, error) {
```
hmlanigan: Needs comment.

neutron/gbp.go
```diff
@@ -0,0 +1,96 @@
...
+		ExpectedStatus: []int{http.StatusCreated},
+	}
+	err := c.client.SendRequest(client.POST, "network", "v2.0", ApiPolicyTargetsV2, &requestData)
+	if err != nil {
+		return nil, errors.Newf(err, "failed to create policy target")
+	}
+	return &resp.PolicyTargetV2, nil
+}
+
+func (c *Client) DeletePolicyTargetV2(ptId string) error {
```
hmlanigan: Needs comment. 

Is it possible for the string to have a space?

hmlanigan (commented):
When will the tests and testservice changes be available?

Can you provide a link to the api doc please?  Searches for it are failing.

jameinel:
As I understand it, the actual API for Group Based Policy is sorely under documented. There is the outline of intent here: https://wiki.openstack.org/wiki/GroupBasedPolicy but there doesn't seem to be concrete API docs.

jameinel:
I do think we'll want to follow up with both tests and better documentation as we work through it, but this is good for now.

$$merge$$

PR #62, opened 2018-03-21
juju bug 1756135, improve handling of service catalog urls 

hmlanigan:
during api version discovery.  Provide coverage to additional OpenStack deployments.
Add Trace to client logging.

Tested against the on-site config, openstack and rackspace, and additional unit tests.

### DIFF
--- a/client/apiversion.go
+++ b/client/apiversion.go
@@ -9,7 +9,6 @@ import (
 	"strconv"
 	"strings"
 
-	gooseerrors "gopkg.in/goose.v2/errors"
 	goosehttp "gopkg.in/goose.v2/http"
 	"gopkg.in/goose.v2/logging"
 )
@@ -67,6 +66,14 @@ func getAPIVersionURL(apiURLVersionInfo *apiURLVersion, requested apiVersion) (s
 		return "", fmt.Errorf("could not find matching URL")
 	}
 	versionURL := apiURLVersionInfo.rootURL
+
+	// https://bugs.launchpad.net/juju/+bug/1756135:
+	// some hrefURL.Path contain more than the version, with
+	// overlap on the apiURLVersionInfo.rootURL
+	if strings.HasPrefix(match, "/"+versionURL.Path) {
+		versionURL.Path = "/"
+	}
+
 	versionURL.Path = path.Join(versionURL.Path, match, apiURLVersionInfo.serviceURLSuffix)
 	return versionURL.String(), nil
 }
@@ -141,6 +148,17 @@ func unmarshallVersion(Versions json.RawMessage) ([]apiVersionInfo, error) {
 // in which case this method will return an empty set of versions in the result
 // structure.
 func (c *authenticatingClient) getAPIVersions(serviceCatalogURL string) (*apiURLVersion, error) {
+	c.apiVersionMu.Lock()
+	defer c.apiVersionMu.Unlock()
+	logger := logging.FromCompat(c.logger)
+
+	// Make sure we haven't already received the version info.
+	// Cache done on serviceCatalogURL, https://<url.Host> is not
+	// guarenteed to be unique.
+	if apiInfo, ok := c.apiURLVersions[serviceCatalogURL]; ok {
+		return apiInfo, nil
+	}
+
 	url, err := url.Parse(serviceCatalogURL)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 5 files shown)


hmlanigan:
[bug 1756135 Controller bootstrap fails against Openstack API behind API Proxy](https://trello.com/c/PY4uFjPM/122-bug-1756135-controller-bootstrap-fails-against-openstack-api-behind-api-proxy)

client/apiversion_test.go
```diff
@@ -118,6 +118,30 @@ func (s *localLiveSuite) makeServiceURLTests() []makeServiceURLTest {
 			success:     false,
 			err:         "no endpoints known for service type: no-such-service",
 		},
+		{
```
nskaggs, line 121: Are these completely representative of everything we saw? I'm thinking of the double version in the URL, like https://10.235.28.252/volumev2/v1/

client/client.go
```diff
@@ -238,14 +238,22 @@ func (c *authenticatingClient) sendAuthRequest(
...
 	if !ok {
 		return "", errors.New("no endpoints known for service type: " + serviceType)
 	}
+
+	defer func() {
+		if returnURL != "" {
+			logger.Tracef("MakeServiceURL: %s",returnURL)
+		}
+	} ()
+
```
nskaggs: Might be nice to include the URL you are falling back to in logger.Warningf("falling back to catalogue service URL").

client/apiversion.go
```diff
@@ -208,27 +240,23 @@ func (c *authenticatingClient) getAPIVersions(serviceCatalogURL string) (*apiURL
...
-			// version endpoint, and will return a status
-			// code of 404. We check for 404 and return an
-			// empty set of versions.
-			logger.Warningf("API version discovery failed: %v", err)
-			c.apiURLVersions[baseURL] = apiURLVersionInfo
-			return apiURLVersionInfo, nil
-		}
-		return nil, err
+	if err := c.sendRequest("GET", getVersionURL, c.Token(), requestData); err != nil {
+		logger.Debugf("API version discovery failed: %v", err)
```
nskaggs: This was warn -- I think we should keep it on warn. We would want every user to see this.

client/apiversion.go
```diff
@@ -67,6 +66,14 @@ func getAPIVersionURL(apiURLVersionInfo *apiURLVersion, requested apiVersion) (s
 		return "", fmt.Errorf("could not find matching URL")
 	}
 	versionURL := apiURLVersionInfo.rootURL
+
+	// https://bugs.launchpad.net/juju/+bug/1756135:
+	// some hrefURL.Path contain more than the version, with
+	// overlap on the apiURLVersionInfo.rootURL
+	if strings.HasPrefix(match, "/"+versionURL.Path) {
+		versionURL.Path = "/"
```
nskaggs: Maybe I'm being verbose, but trace logging for this?

client/apiversion.go
```diff
@@ -151,24 +169,46 @@ func (c *authenticatingClient) getAPIVersions(serviceCatalogURL string) (*apiURL
...
+		// http://y.y.y.y:8774/v2/010ab46135ba414882641f663ec917b6
 		origPathParts = strings.Split(strings.Trim(url.Path, "/"), "/")
 		pathParts = origPathParts
-		if _, err := parseVersion(pathParts[0]); err == nil {
-			pathParts = pathParts[1:]
+		found := false
+		for i, p := range pathParts {
+			if _, err := parseVersion(p); err == nil {
+				found = true
+				if i == 0 {
```
nskaggs, line 198: Debug statements would be useful here for each condition - if we have to do this again, it would be nice to understand how juju parsed it.

nskaggs: Can you also provide an example URL or format for each case?

client/apiversion_test.go
hmlanigan: added.

client/apiversion.go
hmlanigan: done

client/client.go
hmlanigan: We do print out the catalogue near the top of bootstrap... and the url will be printed too if trace is enable.

client/apiversion.go
hmlanigan: added.

client/apiversion.go
hmlanigan: added logger.tracef and additional comments

nskaggs (approved):

nskaggs:
$$merge$$

PR #63, opened 2018-07-11
add client.NewClientTLSConfig

hmlanigan:
add client.NewClientTLSConfig which accepts a *tls.Config for http.  Needed for juju bug:
https://bugs.launchpad.net/juju/+bug/1777897

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -1,6 +1,7 @@
 package client
 
 import (
+	"crypto/tls"
 	"errors"
 	"fmt"
 	"net/url"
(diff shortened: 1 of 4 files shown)


client/client.go
```diff
@@ -177,6 +178,10 @@ func NewNonValidatingClient(creds *identity.Credentials, auth_method identity.Au
 	return newClient(creds, auth_method, goosehttp.NewNonSSLValidating(), logger)
 }
 
+func NewClientTLSConfig(creds *identity.Credentials, auth_method identity.AuthMode, logger logging.CompatLogger, config *tls.Config) AuthenticatingClient {
+	return newClient(creds, auth_method, goosehttp.NewWithTLSConfig(config), logger)
```
SimonRichardson: Strange that `auth_method` is snake case rather than camel case

SimonRichardson (approved):

client/client.go
manadart: And `client_creds`.

manadart (approved):

hmlanigan:
$$merge$$

PR #64, opened 2018-07-25
add method to get Cert for testing with UseTLS = true

hmlanigan:
Added Certificate(servicename) to the openstack test service for testing with UseTLS = true.   Each service in the test services has a different http server, so specify which one the certificate is needed for.

Drive by replace of strings with constants.


SimonRichardson (approved):
LGTM - removing string names is always beneficial :+1:

PR #65, opened 2018-07-27
add cinder. BasicTLSConfig()

hmlanigan:
Allow for a tls.Config to be passed for using a CACert with api calls to the OpenStack by adding cinder. BasicTLSConfig().  Related to: https://bugs.launchpad.net/juju/+bug/1784030

This has been tested by hand, and the current tests, with --live, to verify nothing currently working has been broken have been run.  There is no cinder test service to run against for local tests.   The fix involves returning a routine to be called, so verification of that is difficult.  There is a long term plan to resolve some of these issues, in the mean time, please use with caution.

### DIFF
--- a/cinder/client.go
+++ b/cinder/client.go
@@ -4,6 +4,7 @@
 package cinder
 
 import (
+	"crypto/tls"
 	"fmt"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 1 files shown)


manadart (approved):
This is along the same lines as the prior goose change right?

PR #70, opened 2018-08-09
client: reset a ReqReader if it is necessary to re-authenticate

mhilton:


client/client.go
```diff
@@ -215,8 +217,20 @@ func (c *authenticatingClient) SendRequest(
 	method, svcType, apiVersion, apiCall string,
 	requestData *goosehttp.RequestData,
 ) (err error) {
+	var rs io.ReadSeeker
```
rogpeppe, line 220: If you don't declare err in the return parameters, you won't need this.

internal/io/seekable.go
```diff
@@ -0,0 +1,37 @@
+// Copyright 2018 Canonical Ltd.
+// Licensed under the LGPLv3, see COPYING and COPYING.LESSER file for details.
+
+package io
```
rogpeppe: Could we please not name this the same as the stdlib io package?

internal/io/seekable.go
```diff
@@ -0,0 +1,37 @@
...
+	"fmt"
+	"io"
+
+	"gopkg.in/goose.v2/errors"
+)
+
+// maxBufSize holds the maximum amount of data
+// that may be allocated as a buffer when sending
+// a request when the body is not seekable.
+const maxBufSize = 1024 * 1024 * 1024
```
rogpeppe: This is still quite large... I wonder if we should make buffering scheme pluggable. Something for later.

nova/live_test.go
```diff
@@ -295,7 +295,7 @@ func (s *LiveTests) TestCreateAndDeleteSecurityGroup(c *gc.C) {
 		err = s.nova.DeleteSecurityGroup(group.Id)
 		c.Check(err, gc.IsNil)
 	} else {
-		c.Fatalf("test security group (%d) not found", group.Id)
+		c.Fatalf("test security group (%s) not found", group.Id)
```
rogpeppe, line 298: yay for go vet!

rogpeppe (approved):
LGTM with a couple of minor suggestions, thanks.

client/client.go
mhilton: I'm leaving this as it was already like that.

internal/io/seekable.go
mhilton: done

PR #71, opened 2018-08-10
client: ensure requests are restarted in a race-free way.

mhilton:
HTTP clients can return whilst they are still reading a request
body. Ensure that any request with a body is not restarted until the
previous request has closed it.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -4,7 +4,6 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
-	"io"
 	"net/url"
 	"sort"
 	"strings"
@@ -162,9 +161,9 @@ func newClient(creds *identity.Credentials, auth_method identity.AuthMode, httpC
 		client_creds.URL = client_creds.URL + apiTokens
 	}
 	client := authenticatingClient{
-		creds:                &client_creds,
-		requiredServiceTypes: defaultRequiredServiceTypes,
-		client:               client{logger: logger, httpClient: httpClient},
+		creds:                      &client_creds,
+		requiredServiceTypes:       defaultRequiredServiceTypes,
+		client:                     client{logger: logger, httpClient: httpClient},
 		apiVersionDiscoveryEnabled: true,
 	}
 	client.auth = &client
@@ -217,21 +216,18 @@ func (c *authenticatingClient) SendRequest(
 	method, svcType, apiVersion, apiCall string,
 	requestData *goosehttp.RequestData,
 ) (err error) {
-	var rs io.ReadSeeker
-	rs, err = gooseio.Seekable(requestData.ReqReader, int64(requestData.ReqLength))
-	if err != nil {
-		return
+	if requestData.ReqReader != nil && requestData.GetReqReader == nil {
+		requestData.ReqReader, requestData.GetReqReader = gooseio.MakeGetReqReader(requestData.ReqReader, int64(requestData.ReqLength))
 	}
-	requestData.ReqReader = rs
 	err = c.sendAuthRequest(method, svcType, apiVersion, apiCall, requestData)
 	if gooseerrors.IsUnauthorised(err) {
-		if rs != nil {
-			_, err = rs.Seek(0, 0)
+		c.setToken("")
+		if requestData.GetReqReader != nil {
+			requestData.ReqReader, err = requestData.GetReqReader()
 			if err != nil {
 				return
 			}
 		}
-		c.setToken("")
 		err = c.sendAuthRequest(method, svcType, apiVersion, apiCall, requestData)
 	}
 	return
--- a/http/client.go
+++ b/http/client.go
@@ -66,12 +66,21 @@ type RequestData struct {
 	Params         *url.Values
 	ExpectedStatus []int
 	ReqValue       interface{}
-	// ReqReader is used to read the body of the request.
-	// If it does not implement io.Seeker, the entire
-	// body will be read into memory before sending the request (so
-	// that the request can be retried if needed) otherwise
-	// Seek will be used to rewind the request on retry.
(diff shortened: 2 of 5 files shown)


client/client.go
```diff
@@ -217,21 +216,18 @@ func (c *authenticatingClient) SendRequest(
 	method, svcType, apiVersion, apiCall string,
 	requestData *goosehttp.RequestData,
 ) (err error) {
-	var rs io.ReadSeeker
-	rs, err = gooseio.Seekable(requestData.ReqReader, int64(requestData.ReqLength))
-	if err != nil {
-		return
+	if requestData.ReqReader != nil && requestData.GetReqReader == nil {
+		requestData.ReqReader, requestData.GetReqReader = gooseio.MakeGetReqReader(requestData.ReqReader, int64(requestData.ReqLength))
```
rogpeppe, line 220: I know it did this before too, but side-effecting the contents of requestData is really quite icky. A caller might be reasonably entitled to assume that the ReqReader they put into the RequestData is the same after the call as before.

http/client.go
```diff
@@ -284,25 +296,20 @@ func (c *Client) sendRequest(
...
-				io.Reader
-			}{notifier, reqReader}
-		}
-
-		req, err := http.NewRequest(method, URL, body)
+		req, err := http.NewRequest(method, URL, reqReader)
 		if err != nil {
 			return nil, errors.Newf(err, "failed creating the request %s", URL)
 		}
+		req.GetBody = getReqReader
```
rogpeppe, line 312: For some readers, GetBody is populated automatically; could use that if it does it. *shrug*

internal/gooseio/seekable.go
```diff
@@ -4,34 +4,108 @@
...
+		buf:    make([]byte, size),
 	}
-	reqData := make([]byte, int(length))
-	nrRead, err := io.ReadFull(r, reqData)
-	if err != nil {
-		return nil, errors.Newf(err, "failed reading the request data, read %v of %v bytes", nrRead, length)
+	return br, br.getReqReader
+}
+
+type bufferingReader struct {
```
rogpeppe: It would be nice to have a high level description of what this type does.

internal/gooseio/seekable.go
```diff
@@ -4,34 +4,108 @@
...
+	r      io.Reader
+	closed chan struct{}
+
+	mu     sync.Mutex
+	buf    []byte
+	rp, wp int
+}
+
+func (r *bufferingReader) Read(p []byte) (n int, err error) {
```
rogpeppe: Much as I appreciate a nice bit of slice twiddling, I think it might be possible to avoid it in this case. How about something like this?

	type bufferingReader struct {
		mu  sync.Mutex
		buf bytes.Buffer
		r   io.Reader
	}
	
	func (r *bufferingReader) getReqReader() (io.ReadCloser, error) {
		r.mu.Lock()
		return &unlockingReadCloser{
			mu: &r.mu,
			Reader: io.MultiReader(
				bytes.NewReader(r.buf.Bytes()),
				io.TeeReader(r.r, &r.buf),
			),
		}, nil
	}
	
	type seekingReader struct {
		mu sync.Mutex
		r  io.ReadSeeker
	}
	
	func (r *seekingReader) getReqReader() (io.ReadCloser, error) {
		r.mu.Lock()
		if _, err := r.r.Seek(0, 0); err != nil {
			r.mu.Unlock()
			return nil, err
		}
		return &unlockingReadCloser{
			mu:     &r.mu,
			Reader: r.r,
		}, nil
	}
	
	type unlockingReadCloser struct {
		mu *sync.Mutex
		io.Reader
	}
	
	func (c *unlockingReadCloser) Close() error {
		if c.mu != nil {
			c.mu.Unlock()
			c.mu = nil
		}
		return nil
	}

rogpeppe (approved):
LGTM modulo a couple of thoughts and suggestions, thanks.

internal/gooseio/seekable.go
```diff
@@ -5,33 +5,69 @@ package gooseio
...
+				}
+				return &unlockingReadCloser{
+					mu:     &mu,
+					Reader: rs,
+				}, nil
+			}
 	}
-	if length > maxBufSize {
-		return nil, fmt.Errorf("body of length %d is too large to hold in memory (max %d bytes)", length, maxBufSize)
+	if size > maxBufSize {
```
rogpeppe: This means that we flatly refuse to process any requests that are larger than the requested size. I thought we could do better than that, because it's quite possible that we will receive the error before all of maxBufSize bytes have been read.

The way I was previously envisaging can succeed in that case, I think.

internal/gooseio/seekable.go
```diff
@@ -5,33 +5,69 @@ package gooseio
...
+	var mu sync.Mutex
+	mu.Lock()
+	return &unlockingReadCloser{
+			mu:     &mu,
+			Reader: io.TeeReader(io.LimitReader(r, size), &buf),
+		}, func() (io.ReadCloser, error) {
+			mu.Lock()
+			return &unlockingReadCloser{
+				mu:     &mu,
+				Reader: io.MultiReader(bytes.NewReader(buf.Bytes()), io.TeeReader(io.LimitReader(r, size-int64(buf.Len())), &buf)),
```
rogpeppe: This is quite long. Maybe put the various readers on different lines?
A comment would probably also be appropriate.

rogpeppe (approved):
Looks great, but maybe a final tweak might be good?

internal/gooseio/seekable.go
mhilton: This is still better than the previous behavior which wouldn't even try the first time if it thought it was too long.

PR #72, opened 2018-08-12
Fixed the keystonev3 logic and error messages

vinu2003:
This PR has code changes that fix the error messages returned from goose library related to authentication and also the fix for scope/Project name being added to auth credentials without checking the tenantId or TenantName.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -531,7 +531,7 @@ func (c *authenticatingClient) doAuthenticate() error {
 
 	c.regionServiceURLs = authDetails.RegionServiceURLs
 	if err := c.createServiceURLs(); err != nil {
-		return gooseerrors.Newf(err, "cannot create URLs")
+		return gooseerrors.Newf(err, "cannot create service URLs")
 	}
 	c.apiURLVersions = make(map[string]*apiURLVersion)
 	c.tenantId = authDetails.TenantId
--- a/client/client.go
+++ b/client/client.go
@@ -525,14 +525,14 @@ func (c *authenticatingClient) doAuthenticate() error {
 		err         error
 	)
 	if authDetails, err = c.authMode.Auth(c.creds); err != nil {
-		return gooseerrors.NewUnauthorisedf(err, "", err.Error())
+		return gooseerrors.Newf(err, err.Error())
 	}
 	logger := logging.FromCompat(c.logger)
 	logger.Debugf("auth details: %+v", authDetails)
 
 	c.regionServiceURLs = authDetails.RegionServiceURLs
 	if err := c.createServiceURLs(); err != nil {
-		return gooseerrors.Newf(err, err.Error())
+		return gooseerrors.Newf(err, err.Error(), "cannot create URL")
 	}
 	c.apiURLVersions = make(map[string]*apiURLVersion)
 	c.tenantId = authDetails.TenantId
--- a/client/client.go
+++ b/client/client.go
@@ -17,7 +17,6 @@ import (
 	"gopkg.in/goose.v2/internal/gooseio"
 	"gopkg.in/goose.v2/logging"
 	goosesync "gopkg.in/goose.v2/sync"
-	//"github.com/docker/distribution/registry/api/errcode"
 )
 
 const (
@@ -362,11 +361,11 @@ func (c *authenticatingClient) createServiceURLs() error {
 	}
 	if len(missingServiceTypes) > 0 {
 		if len(possibleRegions) > 0 {
-			return gooseerrors.Newf(nil, "%s\none of these regions may be suitable instead: %s",
+			return fmt.Errorf("%s\none of these regions may be suitable instead: %s",
 				errorPrefix,
 				strings.Join(possibleRegions, ", "))
 		} else {
-			return gooseerrors.Newf(nil, errorPrefix)
+			return fmt.Errorf(errorPrefix)
 		}
 	}
 	c.serviceURLs = serviceURLs
@@ -518,21 +517,21 @@ func (c *authenticatingClient) doAuthenticate() error {
 		return nil
 	}
 	if c.authMode == nil {
-		return gooseerrors.Newf(nil, "Authentication method has not been specified")
+		return fmt.Errorf("Authentication method has not been specified")
 	}
 	var (
 		authDetails *identity.AuthDetails
 		err         error
 	)
 	if authDetails, err = c.authMode.Auth(c.creds); err != nil {
-		return gooseerrors.Newf(err, err.Error())
+		return gooseerrors.Newf(err, "authentication failed")
 	}
 	logger := logging.FromCompat(c.logger)
 	logger.Debugf("auth details: %+v", authDetails)
 
 	c.regionServiceURLs = authDetails.RegionServiceURLs
 	if err := c.createServiceURLs(); err != nil {
-		return gooseerrors.Newf(err, err.Error(), "cannot create URL")
+		return gooseerrors.Newf(err,"cannot create URL")
 	}
 	c.apiURLVersions = make(map[string]*apiURLVersion)
 	c.tenantId = authDetails.TenantId
--- a/client/client.go
+++ b/client/client.go
@@ -17,6 +17,7 @@ import (
 	"gopkg.in/goose.v2/internal/gooseio"
 	"gopkg.in/goose.v2/logging"
 	goosesync "gopkg.in/goose.v2/sync"
+	//"github.com/docker/distribution/registry/api/errcode"
 )
 
 const (
@@ -361,11 +362,11 @@ func (c *authenticatingClient) createServiceURLs() error {
 	}
 	if len(missingServiceTypes) > 0 {
 		if len(possibleRegions) > 0 {
-			return fmt.Errorf("%s\none of these regions may be suitable instead: %s",
+			return gooseerrors.Newf(nil, "%s\none of these regions may be suitable instead: %s",
 				errorPrefix,
 				strings.Join(possibleRegions, ", "))
 		} else {
-			return errors.New(errorPrefix)
+			return gooseerrors.Newf(nil, errorPrefix)
 		}
 	}
 	c.serviceURLs = serviceURLs
@@ -517,29 +518,21 @@ func (c *authenticatingClient) doAuthenticate() error {
 		return nil
 	}
 	if c.authMode == nil {
-		return fmt.Errorf("Authentication method has not been specified")
+		return gooseerrors.Newf(nil, "Authentication method has not been specified")
 	}
 	var (
 		authDetails *identity.AuthDetails
 		err         error
 	)
 	if authDetails, err = c.authMode.Auth(c.creds); err != nil {
-		if gooseerrors.IsUnauthorised(err) {
-			return gooseerrors.NewUnauthorisedf(err, "", "authentication failed: ", c.creds)
-		} else {
-			return gooseerrors.Newf(err, "authentication failed")
-		}
+		return gooseerrors.NewUnauthorisedf(err, "", err.Error())
 	}
 	logger := logging.FromCompat(c.logger)
 	logger.Debugf("auth details: %+v", authDetails)
 
 	c.regionServiceURLs = authDetails.RegionServiceURLs
 	if err := c.createServiceURLs(); err != nil {
-		if gooseerrors.IsUnauthorised(err) {
-			return gooseerrors.NewUnauthorisedf(err, "", "failed to create URLs")
-		} else {
-			return gooseerrors.Newf(err, "cannot create service URLs")
-		}
+		return gooseerrors.Newf(err, err.Error())
 	}
 	c.apiURLVersions = make(map[string]*apiURLVersion)
 	c.tenantId = authDetails.TenantId
--- a/client/client.go
+++ b/client/client.go
@@ -524,14 +524,22 @@ func (c *authenticatingClient) doAuthenticate() error {
 		err         error
 	)
 	if authDetails, err = c.authMode.Auth(c.creds); err != nil {
-		return gooseerrors.Newf(err, "authentication failed")
+		if gooseerrors.IsUnauthorised(err) {
+			return gooseerrors.NewUnauthorisedf(err, "", "authentication failed: ", c.creds)
+		} else {
+			return gooseerrors.Newf(err, "authentication failed")
+		}
 	}
 	logger := logging.FromCompat(c.logger)
 	logger.Debugf("auth details: %+v", authDetails)
 
 	c.regionServiceURLs = authDetails.RegionServiceURLs
 	if err := c.createServiceURLs(); err != nil {
-		return gooseerrors.Newf(err, "cannot create service URLs")
+		if gooseerrors.IsUnauthorised(err) {
+			return gooseerrors.NewUnauthorisedf(err, "", "failed to create URLs")
+		} else {
+			return gooseerrors.Newf(err, "cannot create service URLs")
+		}
 	}
 	c.apiURLVersions = make(map[string]*apiURLVersion)
 	c.tenantId = authDetails.TenantId
--- a/client/client.go
+++ b/client/client.go
@@ -365,7 +365,7 @@ func (c *authenticatingClient) createServiceURLs() error {
 				errorPrefix,
 				strings.Join(possibleRegions, ", "))
 		} else {
-			return fmt.Errorf(errorPrefix)
+			return errors.New(errorPrefix)
 		}
 	}
 	c.serviceURLs = serviceURLs
@@ -531,7 +531,7 @@ func (c *authenticatingClient) doAuthenticate() error {
 
 	c.regionServiceURLs = authDetails.RegionServiceURLs
 	if err := c.createServiceURLs(); err != nil {
-		return gooseerrors.Newf(err,"cannot create URL")
+		return gooseerrors.Newf(err, "cannot create URLs")
 	}
 	c.apiURLVersions = make(map[string]*apiURLVersion)
 	c.tenantId = authDetails.TenantId
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -5,7 +5,6 @@ import (
 	"testing"
 
 	gc "gopkg.in/check.v1"
-
 	"gopkg.in/goose.v2/identity"
 )
 
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	gc "gopkg.in/check.v1"
+
 	"gopkg.in/goose.v2/identity"
 )
 
--- a/client/live_test.go
+++ b/client/live_test.go
@@ -3,6 +3,7 @@ package client_test
 import (
 	gc "gopkg.in/check.v1"
 
+	"fmt"
 	"gopkg.in/goose.v2/client"
 	gooseerrors "gopkg.in/goose.v2/errors"
 	"gopkg.in/goose.v2/identity"
@@ -46,8 +47,8 @@ func (s *LiveTests) TestAuthenticateFail(c *gc.C) {
 	osclient := client.NewClient(&cred, s.authMode, nil)
 	c.Assert(osclient.IsAuthenticated(), gc.Equals, false)
 	err := osclient.Authenticate()
-	isAuthorised := gooseerrors.IsUnauthorised(err)
-	c.Assert(isAuthorised, gc.Equals, true)
+	fmt.Printf("XXXX...... err : %v\n\n", err)
+	c.Assert(err, gc.ErrorMatches, "authentication failed(\n|.)*")
 }
 
 func (s *LiveTests) TestNoUnauthorisedAuthenticateFail(c *gc.C) {
--- a/client/live_test.go
+++ b/client/live_test.go
@@ -50,6 +50,16 @@ func (s *LiveTests) TestAuthenticateFail(c *gc.C) {
 	c.Assert(isAuthorised, gc.Equals, true)
 }
 
+func (s *LiveTests) TestNoUnauthorisedAuthenticateFail(c *gc.C) {
+	cred := *s.cred
+	cred.Region = ""
+	osclient := client.NewClient(&cred, s.authMode, nil)
+	c.Assert(osclient.IsAuthenticated(), gc.Equals, false)
+	err := osclient.Authenticate()
+	isAuthorised := gooseerrors.IsUnauthorised(err)
+	c.Assert(isAuthorised, gc.Equals, false)
+}
+
 func (s *LiveTests) TestAuthenticate(c *gc.C) {
 	cl := client.NewClient(s.cred, s.authMode, nil)
 	err := cl.Authenticate()
--- a/client/live_test.go
+++ b/client/live_test.go
@@ -2,6 +2,7 @@ package client_test
 
 import (
 	gc "gopkg.in/check.v1"
+
 	"gopkg.in/goose.v2/client"
 	gooseerrors "gopkg.in/goose.v2/errors"
 	"gopkg.in/goose.v2/identity"
@@ -54,8 +55,8 @@ func (s *LiveTests) TestNoUnauthorisedAuthenticateFail(c *gc.C) {
 	osclient := client.NewClient(&cred, s.authMode, nil)
 	c.Assert(osclient.IsAuthenticated(), gc.Equals, false)
 	err := osclient.Authenticate()
-	isAuthorised := gooseerrors.IsUnauthorised(err)
-	c.Assert(isAuthorised, gc.Equals, false)
+	isUnAuthorised := gooseerrors.IsUnauthorised(err)
+	c.Assert(isUnAuthorised, gc.Equals, false)
 }
 
 func (s *LiveTests) TestAuthenticate(c *gc.C) {
--- a/client/live_test.go
+++ b/client/live_test.go
@@ -4,6 +4,7 @@ import (
 	gc "gopkg.in/check.v1"
 
 	"gopkg.in/goose.v2/client"
+	gooseerrors "gopkg.in/goose.v2/errors"
 	"gopkg.in/goose.v2/identity"
 )
 
@@ -45,7 +46,8 @@ func (s *LiveTests) TestAuthenticateFail(c *gc.C) {
 	osclient := client.NewClient(&cred, s.authMode, nil)
 	c.Assert(osclient.IsAuthenticated(), gc.Equals, false)
 	err := osclient.Authenticate()
-	c.Assert(err, gc.ErrorMatches, "authentication failed(\n|.)*")
+	isAuthorised := gooseerrors.IsUnauthorised(err)
+	c.Assert(isAuthorised, gc.Equals, true)
 }
 
 func (s *LiveTests) TestAuthenticate(c *gc.C) {
--- a/client/live_test.go
+++ b/client/live_test.go
@@ -2,8 +2,6 @@ package client_test
 
 import (
 	gc "gopkg.in/check.v1"
-
-	"fmt"
 	"gopkg.in/goose.v2/client"
 	gooseerrors "gopkg.in/goose.v2/errors"
 	"gopkg.in/goose.v2/identity"
@@ -47,7 +45,6 @@ func (s *LiveTests) TestAuthenticateFail(c *gc.C) {
 	osclient := client.NewClient(&cred, s.authMode, nil)
 	c.Assert(osclient.IsAuthenticated(), gc.Equals, false)
 	err := osclient.Authenticate()
-	fmt.Printf("XXXX...... err : %v\n\n", err)
 	c.Assert(err, gc.ErrorMatches, "authentication failed(\n|.)*")
 }
 
--- a/client/live_test.go
+++ b/client/live_test.go
@@ -49,14 +49,14 @@ func (s *LiveTests) TestAuthenticateFail(c *gc.C) {
 	c.Assert(err, gc.ErrorMatches, "authentication failed(\n|.)*")
 }
 
-func (s *LiveTests) TestNoUnauthorisedAuthenticateFail(c *gc.C) {
+func (s *LiveTests) TestUnauthorisedAuthenticateFail(c *gc.C) {
 	cred := *s.cred
 	cred.Region = ""
 	osclient := client.NewClient(&cred, s.authMode, nil)
 	c.Assert(osclient.IsAuthenticated(), gc.Equals, false)
 	err := osclient.Authenticate()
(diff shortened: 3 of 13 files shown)


client/client.go
```diff
@@ -524,14 +524,22 @@ func (c *authenticatingClient) doAuthenticate() error {
...
+		}
 	}
 	logger := logging.FromCompat(c.logger)
 	logger.Debugf("auth details: %+v", authDetails)
 
 	c.regionServiceURLs = authDetails.RegionServiceURLs
 	if err := c.createServiceURLs(); err != nil {
-		return gooseerrors.Newf(err, "cannot create service URLs")
+		if gooseerrors.IsUnauthorised(err) {
+			return gooseerrors.NewUnauthorisedf(err, "", "failed to create URLs")
```
wallyworld: Look at the createServiceURLs code, it can never return an UnauthorisedError. It never makes any api calls. It just looks at data structures and returns an error if something is wrong. So all that's needed here is a little extra context added to the error message for user consumption
return gooseerrors.Newf(err, "cannot create service URLs")

client/client.go
```diff
@@ -524,14 +524,22 @@ func (c *authenticatingClient) doAuthenticate() error {
 		err         error
 	)
 	if authDetails, err = c.authMode.Auth(c.creds); err != nil {
-		return gooseerrors.Newf(err, "authentication failed")
+		if gooseerrors.IsUnauthorised(err) {
```
wallyworld: None of this is needed - the key requirement is that an error satisfying IsNotAuthorised is returned if there's an auth problem. You need to look at the Auth() implementation - the underlying issue is for example in v3Auth (and possibly elsewhere) where it does this if auth fails:
return nil, fmt.Errorf("authentication failed")

There's no indication at all that the cause was an auth error.

identity/v3userpass.go
```diff
@@ -103,14 +103,16 @@ func (u *V3UserPass) Auth(creds *Credentials) (*AuthDetails, error) {
 			},
 		},
 	}
-	auth.Auth.Scope = &v3AuthScope{
-		Project: &v3AuthProject{
-			Domain: &v3AuthDomain{
-				Name: projectDomain,
+	if creds.TenantName != "" || creds.TenantID != "" {
```
wallyworld, line 107: Even if both id and name are "", we still need to pass in the project domain I would think? Maybe that's not the case?

wallyworld (commented):
Where are the tests?

identity/v3userpass.go
vinu2003: As discussed, This conditional check is required.
Refer : https://docs.openstack.org/keystone/pike/_modules/keystone/auth/controllers.html

client/client.go
```diff
@@ -517,21 +518,21 @@ func (c *authenticatingClient) doAuthenticate() error {
...
-		return fmt.Errorf("Authentication method has not been specified")
+		return gooseerrors.Newf(nil, "Authentication method has not been specified")
 	}
 	var (
 		authDetails *identity.AuthDetails
 		err         error
 	)
 	if authDetails, err = c.authMode.Auth(c.creds); err != nil {
-		return gooseerrors.Newf(err, "authentication failed")
+		return gooseerrors.NewUnauthorisedf(err, "", err.Error())
```
wallyworld: This is the wrong place - take a look at the underlying code. There can be an error making the json request which is not an auth error but a connectivity error. The place where the actual auth error is determined is the place to use gooseerrors.NewUnauthorisedf

client/client.go
```diff
@@ -517,21 +518,21 @@ func (c *authenticatingClient) doAuthenticate() error {
...
 	}
 	logger := logging.FromCompat(c.logger)
 	logger.Debugf("auth details: %+v", authDetails)
 
 	c.regionServiceURLs = authDetails.RegionServiceURLs
 	if err := c.createServiceURLs(); err != nil {
-		return gooseerrors.Newf(err, "cannot create service URLs")
+		return gooseerrors.Newf(err, err.Error())
```
wallyworld: why remove the annotation?

client/local_test.go
```diff
@@ -222,7 +222,7 @@ func (auth *fakeAuthenticator) Auth(creds *identity.Credentials) (*identity.Auth
 	tooManyCallers := auth.nrCallers > 1
 	auth.mu.Unlock()
 	if tooManyCallers {
-		return nil, fmt.Errorf("Too many callers of Auth function")
+		return nil, errors.NewUnauthorisedf(nil, "", "Too many callers of Auth function")
```
wallyworld: This isn't an auth error, it's a rate limiting error

identity/keystone.go
```diff
@@ -63,14 +62,14 @@ func keystoneAuth(client *goosehttp.Client, auth_data interface{}, URL string) (
 	requestData := goosehttp.RequestData{ReqValue: auth_data, RespValue: &accessWrapper}
 	err := client.JsonRequest("POST", URL, "", &requestData, nil)
 	if err != nil {
-		return nil, err
+		return nil, gooseerrors.NewUnauthorisedf(err, "", "requesting token failed")
```
wallyworld: This isn't an auth error - it's a connectivity error

identity/keystone.go
```diff
@@ -63,14 +62,14 @@ func keystoneAuth(client *goosehttp.Client, auth_data interface{}, URL string) (
...
 	}
 
 	details := &AuthDetails{}
 	access := accessWrapper.Access
 	respToken := access.Token
 	if respToken.Id == "" {
-		return nil, fmt.Errorf("authentication failed")
+		return nil, gooseerrors.NewUnauthorisedf(nil, "", "response token Id is empty")
```
wallyworld: This is correct - this is the bit where we know username/password etc didn't match

identity/legacy.go
```diff
@@ -18,24 +18,24 @@ func (l *Legacy) Auth(creds *Credentials) (*AuthDetails, error) {
...
 	response, err := l.client.Do(request)
 	if err != nil {
-		return nil, err
+		return nil, gooseerrors.NewUnauthorisedf(err, "", "request failed (maybe client policy or network connectivity issue)")
 	}
 	defer response.Body.Close()
 	if response.StatusCode != http.StatusNoContent && response.StatusCode != http.StatusOK {
 		content, _ := ioutil.ReadAll(response.Body)
-		return nil, fmt.Errorf("Failed to Authenticate (code %d %s): %s",
+		return nil, gooseerrors.NewUnauthorisedf(nil, "", "Failed to Authenticate (code %d %s): %s",
```
wallyworld: again here - not an auth error etc

identity/v3userpass.go
wallyworld: You are totally correct

wallyworld (commented):
See embedded comments - we need to be careful to only return auth errors if the error is due to an actually auth check mismatch with the credential. Other things like read io errors, connectivity etc are not auth errors. So why treat them as such as that will just result in juiju printing the error to the user about their credentials being wrong and that's not the issue at all

client/client.go
vinu2003: I can add the add annotation as arg. But, the err.Error() gives all the details to user. I print that out in juju side.

identity/legacy.go
vinu2003: wallyworld: the one @line 32 is auth error but the one @21 is not.

client/client.go
```diff
@@ -17,6 +17,7 @@ import (
 	"gopkg.in/goose.v2/internal/gooseio"
 	"gopkg.in/goose.v2/logging"
 	goosesync "gopkg.in/goose.v2/sync"
+	//"github.com/docker/distribution/registry/api/errcode"
```
wallyworld: delete me

client/client.go
```diff
@@ -517,21 +518,21 @@ func (c *authenticatingClient) doAuthenticate() error {
...
-		return gooseerrors.Newf(err, "authentication failed")
+		return gooseerrors.Newf(err, err.Error())
 	}
 	logger := logging.FromCompat(c.logger)
 	logger.Debugf("auth details: %+v", authDetails)
 
 	c.regionServiceURLs = authDetails.RegionServiceURLs
 	if err := c.createServiceURLs(); err != nil {
-		return gooseerrors.Newf(err, "cannot create service URLs")
+		return gooseerrors.Newf(err, err.Error(), "cannot create URL")
```
wallyworld: The args to Newf() are a format string and associated args to substitute. Is there a teason for err.Error()?
The original code was fine from what I can see.

client/client.go
```diff
@@ -517,21 +518,21 @@ func (c *authenticatingClient) doAuthenticate() error {
...
-		return fmt.Errorf("Authentication method has not been specified")
+		return gooseerrors.Newf(nil, "Authentication method has not been specified")
 	}
 	var (
 		authDetails *identity.AuthDetails
 		err         error
 	)
 	if authDetails, err = c.authMode.Auth(c.creds); err != nil {
-		return gooseerrors.Newf(err, "authentication failed")
+		return gooseerrors.Newf(err, err.Error())
```
wallyworld: The original code is fine here. The arg to Newf is meant to annotate the underlying error to give more context. The err message will already contain the original err string to using err.Error() as the context doesn't make sense.

client/client.go
```diff
@@ -361,11 +362,11 @@ func (c *authenticatingClient) createServiceURLs() error {
 	}
 	if len(missingServiceTypes) > 0 {
 		if len(possibleRegions) > 0 {
-			return fmt.Errorf("%s\none of these regions may be suitable instead: %s",
+			return gooseerrors.Newf(nil, "%s\none of these regions may be suitable instead: %s",
```
wallyworld: I'm not sure that gooseerrors.Newf() is necessary in cases where we pass in nil.
The idea of Newf() is to add extra context to the error without losing the cause. This allows the various gooseerrors.IsFoo() methods to be used.
If we are passing in nil, then fmt.Errorf works just fine.

client/live_test.go
```diff
@@ -45,7 +46,8 @@ func (s *LiveTests) TestAuthenticateFail(c *gc.C) {
 	osclient := client.NewClient(&cred, s.authMode, nil)
 	c.Assert(osclient.IsAuthenticated(), gc.Equals, false)
 	err := osclient.Authenticate()
-	c.Assert(err, gc.ErrorMatches, "authentication failed(\n|.)*")
+	isAuthorised := gooseerrors.IsUnauthorised(err)
+	c.Assert(isAuthorised, gc.Equals, true)
```
wallyworld: We probably need a test where a non auth error like failure to make service urls or whatever is checked so that isAuthorised = false

identity/keystone.go
```diff
@@ -63,14 +62,14 @@ func keystoneAuth(client *goosehttp.Client, auth_data interface{}, URL string) (
 	requestData := goosehttp.RequestData{ReqValue: auth_data, RespValue: &accessWrapper}
 	err := client.JsonRequest("POST", URL, "", &requestData, nil)
 	if err != nil {
-		return nil, err
+		return nil, gooseerrors.Newf(err, "", "requesting token failed")
```
wallyworld: As explained in other comment, revert this

identity/v3userpass.go
wallyworld: Need a test for this

wallyworld (commented):
Looks good but a few things need to be tweaked.
The main issue is use of Newf() - see comments.
Also, there's still some uses of NewUnauthrisedf() which need reverting.
And a test or 2 to add

identity/keystone.go
vinu2003: wallyworld: i dont think this needs to be reverted. we still need the cause. this same code is there in v3userpass.go and legacy.go

client/client.go
```diff
@@ -365,7 +365,7 @@ func (c *authenticatingClient) createServiceURLs() error {
 				errorPrefix,
 				strings.Join(possibleRegions, ", "))
 		} else {
-			return errors.New(errorPrefix)
+			return fmt.Errorf(errorPrefix)
```
wallyworld: just revert to errors.New() in this instance. Only need the Errorf() variant when there are formatting args to worry about.

client/client.go
```diff
@@ -531,7 +531,7 @@ func (c *authenticatingClient) doAuthenticate() error {
 
 	c.regionServiceURLs = authDetails.RegionServiceURLs
 	if err := c.createServiceURLs(); err != nil {
-		return gooseerrors.Newf(err, "cannot create service URLs")
+		return gooseerrors.Newf(err,"cannot create URL")
```
wallyworld: There's a go fmt error here (missing space).
Also, it should be URLs.
If you revert to the original code both issues will be fixed.

client/client_test.go
```diff
@@ -5,7 +5,6 @@ import (
 	"testing"
 
 	gc "gopkg.in/check.v1"
-
```
wallyworld: We need the newline here since goose.v2 is the package we are in

identity/keystone.go
```diff
@@ -63,14 +64,14 @@ func keystoneAuth(client *goosehttp.Client, auth_data interface{}, URL string) (
...
 	}
 
 	details := &AuthDetails{}
 	access := accessWrapper.Access
 	respToken := access.Token
 	if respToken.Id == "" {
-		return nil, fmt.Errorf("authentication failed")
+		return nil, fmt.Errorf("response token Id is empty")
```
wallyworld: What happened to the AuthenticationFailed error? This is precisely where we need this error to be returned.
There should be a failing test with this code.

identity/legacy.go
```diff
@@ -1,10 +1,11 @@
 package identity
 
 import (
-	"fmt"
 	"io/ioutil"
 	"net/http"
 
+	"fmt"
```
wallyworld: move me

identity/legacy.go
```diff
@@ -18,18 +19,18 @@ func (l *Legacy) Auth(creds *Credentials) (*AuthDetails, error) {
 	}
 	request, err := http.NewRequest("GET", creds.URL, nil)
 	if err != nil {
-		return nil, err
+		return nil, gooseerrors.Newf(err, "", "GET request failed")
```
wallyworld: revert me

identity/legacy.go
```diff
@@ -18,18 +19,18 @@ func (l *Legacy) Auth(creds *Credentials) (*AuthDetails, error) {
...
 	}
 	request.Header.Set("X-Auth-User", creds.User)
 	request.Header.Set("X-Auth-Key", creds.Secrets)
 	response, err := l.client.Do(request)
 	if err != nil {
-		return nil, err
+		return nil, gooseerrors.NewUnauthorisedf(err, "", "request failed (maybe client policy or network connectivity issue)")
```
wallyworld: Why is this Unauthorised error? The error message acknowledges that it is probably something else.
You just need to revert this to the original code.

identity/legacy.go
```diff
@@ -18,18 +19,18 @@ func (l *Legacy) Auth(creds *Credentials) (*AuthDetails, error) {
...
 	}
 	defer response.Body.Close()
 	if response.StatusCode != http.StatusNoContent && response.StatusCode != http.StatusOK {
-		content, _ := ioutil.ReadAll(response.Body)
-		return nil, fmt.Errorf("Failed to Authenticate (code %d %s): %s",
+		content, err := ioutil.ReadAll(response.Body)
+		return nil, gooseerrors.NewUnauthorisedf(err, "", "Failed to Authenticate (code %d %s): %s",
```
wallyworld: This isn't unauthrosied either. It's a i/o read error

identity/v3userpass.go
```diff
@@ -1,10 +1,11 @@
 package identity
 
 import (
-	"fmt"
 	"net/http"
 	"time"
 
+	"fmt"
```
wallyworld: move me

identity/v3userpass.go
```diff
@@ -182,11 +185,12 @@ func v3KeystoneAuth(c *goosehttp.Client, v interface{}, url string) (*AuthDetail
 		},
 	}
 	if err := c.JsonRequest("POST", url, "", &req, nil); err != nil {
-		return nil, fmt.Errorf("requesting token: %v", err)
+		fmt.Println("Iam inside Json fail")
```
wallyworld: delete me

identity/v3userpass.go
```diff
@@ -182,11 +185,12 @@ func v3KeystoneAuth(c *goosehttp.Client, v interface{}, url string) (*AuthDetail
...
+		return nil, gooseerrors.Newf(err, "", "requesting token failed")
```
wallyworld: the original code was fine

identity/v3userpass.go
```diff
@@ -182,11 +185,12 @@ func v3KeystoneAuth(c *goosehttp.Client, v interface{}, url string) (*AuthDetail
...
 	}
 	tok := req.RespHeaders.Get("X-Subject-Token")
 	if tok == "" {
-		return nil, fmt.Errorf("authentication failed")
+		return nil, fmt.Errorf("token value associated with X-Subject-Token key is empty")
```
wallyworld: Again, this is exactly the place where we need to return Unauthorised

testservices/identityservice/users.go
```diff
@@ -24,7 +27,7 @@ func (u *Users) addTenant(tenant string) (string, string) {
 	return id, tenant
 }
 
-func (u *Users) AddUser(user, secret, tenant, authDomain string) *UserInfo {
+func (u *Users) AddUser(user, secret, tenant string, authDomain string) *UserInfo {
```
wallyworld: don't need this change

identity/v3userpass_test.go
```diff
@@ -122,3 +122,58 @@ func (s *V3UserPassTestSuite) TestAuthWithCatalog(c *gc.C) {
 	c.Assert(auth.Token, gc.Equals, userInfo.Token)
 	c.Assert(auth.TenantId, gc.Equals, userInfo.TenantId)
 }
+
+func (s *V3UserPassTestSuite) TestAuthFailToDomainwithTenantNameAndTenantID(c *gc.C) {
```
wallyworld: The test name says AuthFail
Are we expecting this test to check for auth fail?

wallyworld (commented):
It looks like there's still some confusion about errors

client/client.go
```diff
@@ -531,7 +531,7 @@ func (c *authenticatingClient) doAuthenticate() error {
 
 	c.regionServiceURLs = authDetails.RegionServiceURLs
 	if err := c.createServiceURLs(); err != nil {
-		return gooseerrors.Newf(err, "cannot create service URLs")
+		return gooseerrors.Newf(err, "cannot create URLs")
```
wallyworld: just revert to what was there before. we need to say "service URLs"

client/live_test.go
```diff
@@ -3,7 +3,9 @@ package client_test
 import (
 	gc "gopkg.in/check.v1"
 
+	"fmt"
```
wallyworld: move me - all std lib imports go to the top

client/live_test.go
```diff
@@ -45,9 +47,20 @@ func (s *LiveTests) TestAuthenticateFail(c *gc.C) {
 	osclient := client.NewClient(&cred, s.authMode, nil)
 	c.Assert(osclient.IsAuthenticated(), gc.Equals, false)
 	err := osclient.Authenticate()
+	fmt.Printf("XXXX...... err : %v\n\n", err)
```
wallyworld: delete me

client/live_test.go
```diff
@@ -45,9 +47,20 @@ func (s *LiveTests) TestAuthenticateFail(c *gc.C) {
...
 	c.Assert(err, gc.ErrorMatches, "authentication failed(\n|.)*")
```
wallyworld: We need gooseerrors.IsUnauthorised(err) check here, should be true

identity/keystone.go
```diff
@@ -63,7 +64,7 @@ func keystoneAuth(client *goosehttp.Client, auth_data interface{}, URL string) (
 	requestData := goosehttp.RequestData{ReqValue: auth_data, RespValue: &accessWrapper}
 	err := client.JsonRequest("POST", URL, "", &requestData, nil)
 	if err != nil {
-		return nil, err
+		return nil, gooseerrors.Newf(err, "", "requesting token failed")
```
wallyworld: No need for the "". The args are standard string formatting args, a format string followed by param values.
You just want gooseerrors.Newf(err, "requesting token failed")

identity/v3userpass.go
```diff
@@ -186,7 +189,7 @@ func v3KeystoneAuth(c *goosehttp.Client, v interface{}, url string) (*AuthDetail
 	}
 	tok := req.RespHeaders.Get("X-Subject-Token")
 	if tok == "" {
-		return nil, fmt.Errorf("authentication failed")
+		return nil, gooseerrors.NewUnauthorisedf(nil, "", "token value associated with X-Subject-Token key is empty")
```
wallyworld: No need for ""
Plus too much internal detail. Maybe just "empty auth token received"

wallyworld (commented):
Almost there.
There's still no test as requested for the v3authScope case when both tenant id and name are empty

client/live_test.go
vinu2003: wallyworld: After reverting back esp the code in legacy.go that IsAuthorised is not true which is correct so we cant make that check here.

vinu2003:
wallyworld: I have the required test without TenantName and TenantId in the v3userpass_test.go Could u please let me know after taking a look at it. I will wait for ur input.
thank u

identity/legacy.go
```diff
@@ -35,7 +36,7 @@ func (l *Legacy) Auth(creds *Credentials) (*AuthDetails, error) {
 	details := &AuthDetails{}
 	details.Token = response.Header.Get("X-Auth-Token")
 	if details.Token == "" {
-		return nil, fmt.Errorf("Did not get valid Token from auth request")
+		return nil, gooseerrors.NewUnauthorisedf(nil, "", "Did not get valid Token from auth request")
```
wallyworld, line 39: remove the extra ""

identity/v3userpass.go
```diff
@@ -186,7 +189,7 @@ func v3KeystoneAuth(c *goosehttp.Client, v interface{}, url string) (*AuthDetail
 	}
 	tok := req.RespHeaders.Get("X-Subject-Token")
 	if tok == "" {
-		return nil, fmt.Errorf("authentication failed")
+		return nil, gooseerrors.NewUnauthorisedf(nil, "token value associated with X-Subject-Token key is empty", "")
```
wallyworld: remove the extra ""

vinu2003: These extra are empty format params which needs to be passed. previously I have added the string in format param but you have asked me to add it in context. However, it still needs to be there. 
If u are mentioning space i have verified that as well now. I beleive it good now.

vinu2003:
hi @wallyworld,
I agree with your point.However, the Auth() returns AuthDetails which get filled in from v3KeyStoneAuth().
The AuthDetails doesnt have ProjectDomainName.
The Logic i have added checks if TenanID or TenantName either of these are set then set ProjectDomainName. If TenantId/Name not there even if credentials has ProjectDomainName it wont be set in auth. 
Inorder to make things visible i added the ProjectDomain Name in AuthDetails struct and validated it.
Kindly let me know your inputs.

client/live_test.go
```diff
@@ -48,6 +48,16 @@ func (s *LiveTests) TestAuthenticateFail(c *gc.C) {
...
 }
 
+func (s *LiveTests) TestNoUnauthorisedAuthenticateFail(c *gc.C) {
+	cred := *s.cred
+	cred.Region = ""
+	osclient := client.NewClient(&cred, s.authMode, nil)
+	c.Assert(osclient.IsAuthenticated(), gc.Equals, false)
+	err := osclient.Authenticate()
+	isAuthorised := gooseerrors.IsUnauthorised(err)
```
howbazaar: This is confusing. You are asking if an error is unauthorised, and then assigning that value into isAuthorised.

Are we expecting a credential without a region to return an unauthorised error?

client/live_test.go
```diff
@@ -2,8 +2,8 @@ package client_test
 
 import (
 	gc "gopkg.in/check.v1"
-
```
howbazaar: This blank line is needed back in.

As a general rule, we separate imports into three different blocks: system packages, other package, packages from this repo.

client/live_test.go
```diff
@@ -48,6 +48,16 @@ func (s *LiveTests) TestAuthenticateFail(c *gc.C) {
 	c.Assert(err, gc.ErrorMatches, "authentication failed(\n|.)*")
 }
 
+func (s *LiveTests) TestNoUnauthorisedAuthenticateFail(c *gc.C) {
```
howbazaar: What are we actually testing here? The test method makes no sense to me.

identity/legacy.go
```diff
@@ -35,7 +36,7 @@ func (l *Legacy) Auth(creds *Credentials) (*AuthDetails, error) {
 	details := &AuthDetails{}
 	details.Token = response.Header.Get("X-Auth-Token")
 	if details.Token == "" {
-		return nil, fmt.Errorf("Did not get valid Token from auth request")
+		return nil, gooseerrors.NewUnauthorisedf(nil, "Did not get valid Token from auth request", "")
```
howbazaar: The trailing "" here isn't needed. The methods that end in 'f' generally follow the formatting rules of things like Printf,  Sprintf, or Fprintf.  There may be extra args, in this case the first arg being the error cause, then there is the format string, which can include the %s, %v type replacements, then the last args are the values to be formatted into the format string. Here we don't have any formmating values, so the trailing arg here isn't needed.

identity/v3userpass.go
howbazaar: What happens if neither is set?

identity/v3userpass.go
```diff
@@ -186,7 +189,7 @@ func v3KeystoneAuth(c *goosehttp.Client, v interface{}, url string) (*AuthDetail
 	}
 	tok := req.RespHeaders.Get("X-Subject-Token")
 	if tok == "" {
-		return nil, fmt.Errorf("authentication failed")
+		return nil, gooseerrors.NewUnauthorisedf(nil, "empty auth token received.", "")
```
howbazaar: As mentioned above, the tailing empty string isn't needed.

identity/v3userpass_test.go
```diff
@@ -122,3 +122,59 @@ func (s *V3UserPassTestSuite) TestAuthWithCatalog(c *gc.C) {
...
+func (s *V3UserPassTestSuite) TestAuthToProjectDomainWithoutTenantNameAndTenantID(c *gc.C) {
+	service := identityservice.NewV3UserPass()
+	service.SetupHTTP(s.Mux)
+	userInfo := service.AddUser("joe-user", "secrets", "", "default")
+	var l Authenticator = &V3UserPass{}
+	creds := Credentials{
+		User:          "joe-user",
+		URL:           s.Server.URL + "/v3/auth/tokens",
+		Secrets:       "secrets",
+		ProjectDomain: "project-domain",
```
howbazaar, line 174: Should this be "project-domain"? Looks like it could be copy and paste error from above.

howbazaar (commented):
Generally looking like the right direction. Mostly minor things.

We have two main update in the code where we are now returning unauthorised errors, but only one test that checks for unauthorised errors. Are we missing another?

client/live_test.go
vinu2003: The func name  needs to be changed. we make sure it is not authorisation issue based on some invalid username/password combination. 
we categorise authorisation issues from other issues like connectivity issues etc..;. Region being empty is not authorisation problem it is setup issue. So that user is provided with correct details when there is an issue.

client/live_test.go
vinu2003: The Test name is changed.

identity/legacy.go
vinu2003: actually its not trailing "". it is format and after which the args come in. Initially wallyworld asked me to get this message inside the context part where as i have added it in format part and if required any args. Thats correct way to do it.
I spoke to him morning and he is fine i changed it. the next commit will have it.

identity/v3userpass.go
vinu2003: the ProjectDomain Name wont get set. Which should not be even if the user provides while 'juju add-credentials'. Becuase openstack will throw an error if this param is set but either of TenantName  or TenantId not set. I have faced this issue during smoke test and added this logic. Please refer to above link.

identity/v3userpass.go
vinu2003: move the message to format. the context can be empty.

identity/v3userpass_test.go
vinu2003: i didnt copy. i just gave a simple name to test. something relative. 
Actually we have domain/userDomain/ProjectDomain.

client/live_test.go
vinu2003: done

identity/v3userpass_test.go
```diff
@@ -122,3 +123,161 @@ func (s *V3UserPassTestSuite) TestAuthWithCatalog(c *gc.C) {
...
+		Secrets:    "secrets",
+		TenantName: "tenant",
+		TenantID:   "tenantID",
+		Domain:     "domain",
+	}
+	auth, err := l.Auth(&creds)
+	c.Assert(err, gc.IsNil)
+	c.Assert(auth.Token, gc.Equals, userInfo.Token)
+	c.Assert(auth.Domain, gc.Equals, "domain")
+	c.Assert(auth.TenantName, gc.Equals, "")
```
howbazaar: Shouldn't the TenantName and TenantId be set?

identity/v3userpass_test.go
```diff
@@ -83,6 +83,7 @@ func (s *V3UserPassTestSuite) TestAuthToTenantNameAndTenantID(c *gc.C) {
 	c.Assert(auth.Token, gc.Equals, userInfo.Token)
 	c.Assert(auth.TenantName, gc.Equals, userInfo.TenantName)
 	c.Assert(auth.TenantId, gc.Equals, userInfo.TenantId)
+	//fmt.Printf("XXXX.... userinfo.TenantName : %v \n", userInfo.TenantName)
```
howbazaar: Remove this line before landing.

howbazaar (approved):
I think you are good to land this now, just remove the commented out line and you are good to merge.

vinu2003:
[openstack auth bugs](https://trello.com/c/SxLOXDFz/373-openstack-auth-bugs)

vinu2003:
[openstack auth bugs](https://trello.com/c/SxLOXDFz/373-openstack-auth-bugs)

PR #73, opened 2018-08-14
client: ensure requests are restarted in a race-free way.

rogpeppe:
HTTP clients can return whilst they are still reading a request body. Ensure that any request with a body is not restarted until the request has closed it.

Note: this is an update to #71 by @mhilton - the changes I've added are in a separate commit in this PR.


mhemmings (approved):
Neat, LGTM

PR #75, opened 2019-06-06
Added block_device_mapping_v2 to RunServer API

hpidcock:
Added BlockDeviceMappings field to RunServerOpts to allow for fine grained control over block devices when creating a server on OpenStack.
Also changed ImageRef to be optional, as this should only be specified when using legacy OpenStack create-server behaviour.

See https://developer.openstack.org/api-ref/compute/?expanded=create-server-detail on the block_device_mapping_v2 schema.

### DIFF
--- a/nova/json.go
+++ b/nova/json.go
@@ -112,6 +112,10 @@ func (entity Entity) MarshalJSON() ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


babbageclunk (approved):
Looks fantastic!

wallyworld:
$$merge$$

PR #80, opened 2020-03-03
Port tags

SimonRichardson:
In order to identify ports once created and to be able to filter
them using list ports, it's ideal to place tags on each port.

The code updates the port struct to add the tag and then goes
through the test harness to expect the tag field in the tests.

#### Reference

See the creation of ports documentation:
https://docs.openstack.org/api-ref/network/v2/index.html?expanded=create-port-detail#ports


hmlanigan (approved):
LGTM from an OpenStack point of view.

PR #81, opened 2020-03-10
Deprecated API Requests.

SimonRichardson:
When looking over the documentation to add new API requests, some of our
existing API calls are using deprecated methods. The following changes
calls out the methods and sets them aside in a deprecated file, until we
have a plan to create a new v3 goose.

The plan would entail if we remove the old calls and provide no
alternative or if we add the new calls to allow migration forward for
others.

Either way, the first step of calling out what shouldn't be used is
probably a good path forward.


hmlanigan (approved):
LGTM, I double checked deprecations via official docs and luckily there is no leaking of the deprecated structs back into nova.go

SimonRichardson:
$$merge$$

PR #82, opened 2020-03-12
OS interfaces

SimonRichardson:
The following adds the `/servers/{server_id}/os-interface` endpoint
for listing all ports associated with a server.

Added subsequent tests to the stubs to ensure that we test the getting
the expected interfaces back from the stub server.

The is required to fulfill the need from Juju to be able to get all the
ports of a given server, to then be able to delete them when destroying
a controller.

#### Refrence

The documentation for this is here:
https://docs.openstack.org/api-ref/compute/?expanded=list-port-interfaces-detail


nova/live_test.go
```diff
@@ -747,3 +747,30 @@ func (s *LiveTests) TestCreateServerBlockDeviceMappingVolume(c *gc.C) {
...
+		Name:             "inst-os-interface",
+		FlavorId:         s.testFlavorId,
+		ImageId:          s.testImageId,
+		AvailabilityZone: s.testAvailabilityZone,
+		Networks: []nova.ServerNetworks{{
+			NetworkId: s.testNetwork,
+		}},
+	})
+	c.Assert(err, gc.IsNil)
+	defer s.nova.DeleteServer(entity.Id)
```
hmlanigan: ignored returned error

nova/nova.go
```diff
@@ -590,3 +590,38 @@ func (c *Client) SetServerMetadata(serverId string, metadata map[string]string)
...
+
+// OSInterface represents an interface attachment to a server.
+type OSInterface struct {
+	FixedIPs   []PortFixedIP `json:"fixed_ips,omitempty"`
+	IPAddress  string        `json:"ip_address"`
+	MacAddress string        `json:"mac_addr,omitempty"`
+	NetID      string        `json:"net_id,omitempty"`
+	PortID     string        `json:"port_id,omitempty"`
+	PortState  string        `json:"port_state,omitempty"`
+}
```
hmlanigan, line 610: Just for completeness.

```suggestion
	Tag         string        `json:"tag,omitempty"`
}
```

nova/nova.go
```diff
@@ -590,3 +590,38 @@ func (c *Client) SetServerMetadata(serverId string, metadata map[string]string)
...
+
+// ListOSInterfaces lists all the os-interfaces (port interfaces) associated
+// with a given server.
+// https://docs.openstack.org/api-ref/compute/?expanded=list-port-interfaces-detail
+func (c *Client) ListOSInterfaces(serverId string) ([]OSInterface, error) {
```
hmlanigan, line 615: Would it be useful to add call to encapsulate: Show Port Interface Details?  /servers/{server_id}/os-interface/{port_id}.  It provides slightly more data.

nova/nova.go
```diff
@@ -590,3 +590,38 @@ func (c *Client) SetServerMetadata(serverId string, metadata map[string]string)
...
+	var resp struct {
+		InterfaceAttachments []OSInterface `json:"interfaceAttachments"`
+	}
+	requestData := goosehttp.RequestData{
+		RespValue: &resp,
+	}
```
hmlanigan, line 622:
```suggestion
            ExpectedStatus: []int{http.StatusOK},
	}
```

hmlanigan (approved):
A few suggestions, but overall LGTM

nova/nova.go
SimonRichardson: I did think about doing that, I'll make a follow-up card to look at that in another PR.

PR #83, opened 2020-03-27
Add MACAddress property to PortV2

manadart:
The port MAC address is required to in order to add cloud-init network configuration to servers that ports are attached to.


SimonRichardson (approved):
:+1:

PR #85, opened 2020-04-01
Allow overloading of headers.

SimonRichardson:
The following changes expose the http headers to us as a library user
can alter them if required. This happens to be the case for juju, where
we want to ensure that some headers are not set.

This is a halfway house as the real fix is to create a factor method for
handling the different client constructors. Then you can pass in a
header method to the factory to then create them. Unfortunately, we're
late in the cycle and doing that at this late stage is _very_ risky. So
let's take the more moderate approach and come back to this for V3 of
goose.

---

It should be noted this is backwards compatible with existing code, as
we're using optional method parameters.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -93,8 +93,26 @@ type AuthenticatingClient interface {
 	IdentityAuthOptions() (identity.AuthOptions, error)
 }
 
-// A single http client is shared between all Goose clients.
-var sharedHttpClient = goosehttp.New()
+// Option allows the adaptation of a client given new options.
+type Option func(*options)
+
+type options struct {
+	httpHeadersFunc goosehttp.HeadersFunc
+}
+
+// WithHTTPHeadersFunc allows passing in a new HTTP headers func for the client
+// to execute for each request.
+func WithHTTPHeadersFunc(httpHeadersFunc goosehttp.HeadersFunc) Option {
+	return func(options *options) {
+		options.httpHeadersFunc = httpHeadersFunc
+	}
+}
+
+func newOptions() *options {
+	return &options{
+		httpHeadersFunc: goosehttp.DefaultHeaders,
+	}
+}
 
 // This client sends requests without authenticating.
 type client struct {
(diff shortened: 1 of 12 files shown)


client/client.go
```diff
@@ -169,28 +240,19 @@ func newClient(creds *identity.Credentials, auth_method identity.AuthMode, httpC
...
-}
-
-func NewNonValidatingClient(creds *identity.Credentials, authMethod identity.AuthMode, logger logging.CompatLogger) AuthenticatingClient {
-	return newClient(creds, authMethod, goosehttp.NewNonSSLValidating(), logger)
-}
-
-func NewClientTLSConfig(creds *identity.Credentials, authMethod identity.AuthMode, logger logging.CompatLogger, config *tls.Config) AuthenticatingClient {
-	return newClient(creds, authMethod, goosehttp.NewWithTLSConfig(config), logger)
-}
-
```
SimonRichardson, line 193: I've just congregated them together with the other constructors to highlight the problem at hand.

neutron/clientheaders.go
```diff
@@ -0,0 +1,63 @@
...
+	// Officially we should also take into account the method, as we should not
+	// be applying this to every request.
+	if payloadExists {
+		headers.Add("Content-Type", contentType)
+	}
+
+	// According to the REST specs, the following should be considered the
+	// correct implementation of requests. Openstack implementation follow
+	// these specs and will return errors if certain headers are pressent.
+	//
```
SimonRichardson, line 33: We explicitly don't follow this, because of the `Content-Type`, which used the `payloadExists` rather than the method. Instead it should really be (pseudo code):

```go
switch method {
case "POST", "PUT", "PATCH":
	if payloadExists {
		headers.Add("Content-Type", contentType)
	}
default:
	// We don't care here.
}
```

neutron/neutron.go
```diff
@@ -251,12 +253,14 @@ func (c *Client) DeleteFloatingIPV2(ipId string) error {
...
 type PortV2 struct {
+	AdminStateUp        bool             `json:"admin_state_up,omitempty"`
 	Description         string           `json:"description,omitempty"`
 	DeviceId            string           `json:"device_id,omitempty"`
+	DeviceOwner         string           `json:"device_owner,omitempty"`
 	FixedIPs            []PortFixedIPsV2 `json:"fixed_ips,omitempty"`
 	Id                  string           `json:"id,omitempty"`
 	Name                string           `json:"name,omitempty"`
-	NetworkId           string           `json:"network_id"`
+	NetworkId           string           `json:"network_id,omitempty"`
```
SimonRichardson, line 263: If you try and update the Port with an empty `NetworkId` it will fail with an error. Instead, we have to omit empty even for other methods. V3 of the library should look into Port types for method verb requests.

http/client.go
```diff
@@ -37,8 +37,27 @@ type HttpClient interface {
 	PostForm(url string, data url.Values) (resp *http.Response, err error)
 }
 
+type Option func(*options)
```
manadart, line 44: Perhaps a note on the independence of this type from the one in _client/client.go_

manadart (approved):
Good change in the context of what we're working with. I guess we can't justify a bigger overhaul right now.

PR #86, opened 2020-06-25
Allow delete non-existent security group.

howbazaar:
Getting a not found error when deleting something should be a silent success.

Addresses https://bugs.launchpad.net/juju/+bug/1864564


howbazaar:
I couldn't find any existing tests for this.

SimonRichardson (changes requested):
I don't think this right, especially for a client. Instead, we should return an `errors.NotFound` and then correctly handle the `NotFound` error inside the Openstack firewaller.

SimonRichardson:
See: https://github.com/juju/juju/pull/11756

howbazaar:
Fair call.

PR #87, opened 2020-06-30
Add ListAvailabilityZones API to volume endpoint

wallyworld:
Add the ListVolumeAvailabilityZones() API to the cinder client


hmlanigan (approved):
LTGM, we should continue this and return the AvailabilityZoneState for compute as well.

PR #90, opened 2021-06-04
HTTP: Allow passing http client

SimonRichardson:
Passing in the http client, allows us to ensure the same client is used
throughout the lifecycle of a goose query. The previous implementations
either used the default http client in the stdlib or it created a new
http.client depending on circumstances of how you create the goose http
client.

The code breaks compatibility by providing a v3 of this library.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -4,6 +4,7 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
+	"net/http"
 	"net/url"
 	"sort"
 	"strings"
@@ -100,7 +101,9 @@ type AuthenticatingClient interface {
 type Option func(*options)
 
 type options struct {
-	httpHeadersFunc goosehttp.HeadersFunc
+	httpHeadersFunc    goosehttp.HeadersFunc
+	httpClient         *http.Client
+	insecureHTTPClient *http.Client
 }
 
 // WithHTTPHeadersFunc allows passing in a new HTTP headers func for the client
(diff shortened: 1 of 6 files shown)


http/client.go
```diff
@@ -144,53 +152,47 @@ func New(options ...Option) *Client {
...
 	}
-
-	return &Client{
-		Client:          defaultClient,
-		headersFunc:     opts.headersFunc,
-		maxSendAttempts: MaxSendAttempts,
+	if client.Transport == nil {
+		client.Transport = &http.Transport{}
 	}
+	if t, ok := client.Transport.(*http.Transport); ok {
```
hmlanigan: We should return an error if ok false. Otherwise you'd get a client without a TLS config and not know it.

client/client.go
```diff
@@ -226,7 +265,19 @@ func NewClientTLSConfig(creds *identity.Credentials, authMethod identity.AuthMod
 		option(opts)
 	}
 
-	return newClient(creds, authMethod, goosehttp.NewWithTLSConfig(config, goosehttp.WithHeadersFunc(opts.httpHeadersFunc)), logger)
+	client := *opts.httpClient
+	if client.Transport == nil {
+		client.Transport = &http.Transport{}
+	}
+	if t, ok := client.Transport.(*http.Transport); ok {
+		t.TLSClientConfig = config
```
hmlanigan, line 279: We should return an error if ok false. Otherwise you'd get a client without a TLS config and not know it.

http/client.go
```diff
@@ -144,53 +152,47 @@ func New(options ...Option) *Client {
...
-		Client:          *http.DefaultClient,
+		Client:          *opts.httpClient,
 		headersFunc:     opts.headersFunc,
 		maxSendAttempts: MaxSendAttempts,
 	}
 }
 
 // NewNonSSLValidating creates a new goose http *Client skipping SSL validation.
+// DEPRECATED (stickupkid): No longer used internally, left to ensure we don't
+// break compatibility.
```
hmlanigan: Can we move to a v3, break compatibility and clean more up?  Especially with a few client methods needing to return an error.

hmlanigan (changes requested):

hmlanigan:
@SimonRichardson description of the PR needs updates related to v3

hmlanigan:
@SimonRichardson the description needs updating about v3.

hmlanigan (approved):
Great!  LGTM.

PR #91, opened 2021-06-04
Linter: Introduce static analysis

SimonRichardson:
Following on from #90 it was noticed that there was some dead code, so
to improve the codebase a static-analysis tool was added and run. The
subsequent changes are the fallout from running the tool.


test.py
```diff
@@ -134,40 +65,13 @@ def run_cmd(cmd):
...
 
 
-def run_go_fmt(opts):
-    return run_cmd(['go', 'fmt', './...'])
-
-
-def run_go_build(opts):
-    return run_cmd(['go', 'build', './...'])
-
-
```
hmlanigan, line 144: can we add these targets to the Makefile?

A default make target that runs static-analysis and go build would be nice.  The default make target as static-analysis is unusual.

hmlanigan (approved):
Thank you for adding this.  A nice improvement.  One suggestion for the Makefile.

test.py
SimonRichardson: I don't think we need the `build`, as there is nothing to build (I think the python code was copied from other projects). `go fmt` is an IDE thing, that is picked and checked via the static analysis now, so is redundant as well.

PR #92, opened 2021-06-08
Version: Move to v3 from v2

SimonRichardson:
The following is a very mechanical change that moves from v2 to v3.

See: https://gopkg.in/goose.v3


manadart (approved):
All good.

SimonRichardson:
$$merge$$

PR #94, opened 2021-06-29
Logging: Default logger

SimonRichardson:
Remove the loggo dependency from the non-test code base. You don't need
a loggo dependency directly for logging, instead you just need a Debugf
method. The fact that a direct dependency on this was required makes it
very difficult to iterate the loggo logger in the new modern world of go
mod.

Remove the comment about just depending on loggo, I really think that's
the wrong thing to do.

### DIFF
--- a/logging/logger.go
+++ b/logging/logger.go
@@ -5,16 +5,12 @@ package logging
 
 import (
 	"log"
-
-	"github.com/juju/loggo"
 )
 
 // CompatLogger is a minimal logging interface that may be provided
 // when constructing a goose Client to log requests and responses,
 // retaining compatibility with the old *log.Logger that was
 // previously depended upon directly.
-//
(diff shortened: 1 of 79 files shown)


logging/logger.go
```diff
@@ -29,17 +25,23 @@ type Logger interface {
...
+}
+
+// DebugLoggerAdapater is a logger that may be provided when constructing
 // a goose Client to log requests and responses. Users must
 // provide a CompatLogger, which will be upgraded to Logger
 // if provided.
-type LoggoLogger struct {
-	loggo.Logger
+type DebugLoggerAdapater struct {
+	Logger DebugLogger
```
SimonRichardson, line 39: This is the real fix.

jameinel (approved):

hmlanigan (approved):
LTGM - wish we had time to take it further

SimonRichardson:
$$merge$$

PR #95, opened 2021-11-04
Use network names in server.Addresses instead of fake names

hmlanigan:
Update how the nova test service creates ServerDetail.  The names in Addresses should be actual network names that exist for juju testing.


hmlanigan:
Required for https://github.com/juju/juju/pull/13478

manadart (approved):
Yes, I see.

hmlanigan:
$$merge$$

PR #96, opened 2022-04-19
Do not combine unauthorised and forbidden error codes

wallyworld:
Both 401 and 403 were being interpreted as an UnauthorisedError. However, 403 can also be used for quota exceeded. The result is that Juju is invalidating credentials when a quota is exceed, instead of simply recording the error and allowing for a retry after the quota issue is fixed.

https://bugs.launchpad.net/juju/+bug/1966375


ycliuhw (approved):
LGTM ty

PR #100, opened 2025-07-04
feat: support to create tags and filter security groups

adisazhar123:
This PR adds a new API to:
#### 1. Add tags to a resource -- in this case it's a security group.
This is done through the `PUT /v2.0/{resourceType}/{resourceId}/tags` endpoint where `resourceType` and `resourceId` are `security-groups` and the identifier, respectively. This endpoint replaces a group's existing tags with the given tags in the request body.

Quoting from Openstack docs ([ref](https://docs.openstack.org/api-ref/network/v2/#standard-attributes-tag-extension)), `resourceType` may be:

> This includes networks, ports, subnets, subnet pools, floating IPs, routers, security groups, security group rules, QoS policies and trunks.

One thing I have noticed is that Openstack documentation mentions that a `POST` method exists to create tags for a resource but upon testing, I get error:
```json
{
	"NeutronError": {
		"type": "HTTPNotFound",
		"message": "not supported",
		"detail": ""
	}
}
```

I then tested using Openstack CLI to create a security group with tags, and they use the `PUT` method as opposed of `POST`. For consistency, I have moved forward to using the `PUT` method.

```bash
openstack security group create <group-name> --tag <tag-name> --debug
```

#### 2. Fetch security groups by tags

This is done through the `GET /v2.0/security-groups` by passing a `tags` query parameter. Multiple tags should be separated by a comma.

Card: https://warthogs.atlassian.net/browse/JUJU-8225

### DIFF
--- a/neutron/neutron.go
+++ b/neutron/neutron.go
@@ -363,6 +363,7 @@ type SecurityGroupV2 struct {
 	Id          string                `json:"id"`
 	Name        string                `json:"name"`
 	Description string                `json:"description"`
+	Tags        []string              `json:"tags"`
 }
 
 type ListSecurityGroupsV2Query struct {
@@ -376,7 +377,7 @@ func (c *Client) ListSecurityGroupsV2(query ListSecurityGroupsV2Query) ([]Securi
 		Groups []SecurityGroupV2 `json:"security_groups"`
 	}
 	requestData := goosehttp.RequestData{RespValue: &resp}
-	endpoint := ApiApplicationPolicyGroupsV2
+	endpoint := ApiSecurityGroupsV2
 
 	if len(query.Tags) > 0 {
 		endpoint = fmt.Sprintf("%s?tags=%s", endpoint, url.QueryEscape(strings.Join(query.Tags, ",")))
@@ -550,16 +551,15 @@ func (c *Client) DeleteSecurityGroupRuleV2(ruleId string) error {
 	return err
 }
 
-// CreateTags creates multiple tags for a resource.
-func (c *Client) CreateTags(resourceType string, resourceId string, tags []string) error {
+// ReplaceAllTags Replaces all tags on the resource.
+func (c *Client) ReplaceAllTags(resourceType string, resourceId string, tags []string) error {
 	var req struct {
 		Tags []string `json:"tags"`
 	}
 	req.Tags = tags
-
 	endpoint := fmt.Sprintf("%s/%s/tags", resourceType, resourceId)
-	requestData := goosehttp.RequestData{ExpectedStatus: []int{http.StatusOK}}
-	err := c.client.SendRequest(client.POST, "tags", "v2.0", endpoint, &requestData)
+	requestData := goosehttp.RequestData{ReqValue: req, ExpectedStatus: []int{http.StatusOK}}
+	err := c.client.SendRequest(client.PUT, "network", "v2.0", endpoint, &requestData)
 	if err != nil {
 		err = errors.Newf(err, "failed to create tags %v for resource type %s", req.Tags, resourceType)
 	}
--- a/neutron/neutron.go
+++ b/neutron/neutron.go
@@ -410,8 +410,25 @@ func (c *Client) SecurityGroupByNameV2(name string) ([]SecurityGroupV2, error) {
 	return resp.Groups, nil
 }
 
+func (c *Client) addTagsToSecurityGroup(group *SecurityGroupV2, tags []string) error {
+	if len(tags) > 0 {
+		err := c.ReplaceAllTags("security-groups", group.Id, tags)
+		// we have to roll back, delete the security group
+		if err != nil {
+			if deleteErr := c.DeleteSecurityGroupV2(group.Id); deleteErr != nil {
+				return errors.Newf(deleteErr, "failed to roll back security group with id: %s", group.Id)
+			}
+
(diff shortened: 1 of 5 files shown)


neutron/neutron.go
```diff
@@ -362,15 +363,27 @@ type SecurityGroupV2 struct {
 	Id          string                `json:"id"`
 	Name        string                `json:"name"`
 	Description string                `json:"description"`
+	Tags        []string              `json:"tags"`
+}
+
+type ListSecurityGroupsV2Query struct {
+	// Security groups that match all entries in `Tags` will be returned.
```
wallyworld: Mention that Tags is optional?
eg
`// Security groups that match all entries in `Tags` (if any are specified) will be returned.`

neutron/neutron.go
```diff
@@ -537,3 +550,19 @@ func (c *Client) DeleteSecurityGroupRuleV2(ruleId string) error {
 	}
 	return err
 }
+
+// ReplaceAllTags Replaces all tags on the resource.
```
wallyworld: `on the specified network resource`

(Just making it explicit that it's only network resources that we are tagging, not resources in general)

wallyworld: Can we unexport this can call it as part of the CreateSecurityGroup call to goose. There's no need for the SDK client to have to make the 2 calls. SO it comceptually looks like the CLI and also like the create vm api call. And we'll then also need to support rollback - deleting the sec group if the tag call fails. But this would be handled internally here in the SDK, not by the caller.

wallyworld (commented):
Thank you. See comments for a suggested change.

neutron/neutron.go
adisazhar123: Done, thanks.

neutron/neutron.go
adisazhar123: > Just making it explicit

Yup, I agree.

> Can we unexport this can call it as part of the CreateSecurityGroup call to goose

This is a good point. I've made changes so that the sec group and tags creation are atomic. To test the rollback, I've added a marker tag to indicate that the test double should fail if it sees that tag. Not sure if there's a better way, but open to suggestions.

adisazhar123:
@wallyworld thanks for the feedback. I have addressed your comments. Need your help to review again.

neutron/neutron.go
```diff
@@ -362,15 +363,27 @@ type SecurityGroupV2 struct {
 	Id          string                `json:"id"`
 	Name        string                `json:"name"`
 	Description string                `json:"description"`
+	Tags        []string              `json:"tags"`
+}
+
+type ListSecurityGroupsV2Query struct {
```
wallyworld, line 371: Comment the struct itself

neutron/neutron.go
```diff
@@ -397,8 +410,25 @@ func (c *Client) SecurityGroupByNameV2(name string) ([]SecurityGroupV2, error) {
 	return resp.Groups, nil
 }
 
+func (c *Client) addTagsToSecurityGroup(group *SecurityGroupV2, tags []string) error {
+	if len(tags) > 0 {
```
wallyworld: Better as if len(tags) == 0 return early. Reduces indentation etc

neutron/neutron.go
```diff
@@ -420,6 +450,12 @@ func (c *Client) CreateSecurityGroupV2(name, description string) (*SecurityGroup
 	if err != nil {
 		return nil, errors.Newf(err, "failed to create a security group with name: %s", name)
 	}
+
+	err = c.addTagsToSecurityGroup(&resp.SecurityGroup, tags)
```
wallyworld: If this fails we need to attempt to rollback the creation of the group

wallyworld (commented):
Just need to rollback the group creation if the tags fail

neutron/neutron.go
adisazhar123: Done, thanks.

neutron/neutron.go
adisazhar123: Done, thanks. FYI moved this to `CreateSecurityGroupV2`.

neutron/neutron.go
adisazhar123: The rollback was inside `addTagsToSecurityGroup` but I have extracted it out to be in the same level as `CreateSecurityGroupV2`

neutron/neutron.go
```diff
@@ -420,6 +434,24 @@ func (c *Client) CreateSecurityGroupV2(name, description string) (*SecurityGroup
...
+	if len(tags) == 0 {
+		return &resp.SecurityGroup, nil
+	}
+
+	// Create the tags for the group.
+	err = c.ReplaceAllTags("security-groups", resp.SecurityGroup.Id, tags)
+	// If this fails, we have to roll back by deleting the security group.
+	if err != nil {
+		if deleteErr := c.DeleteSecurityGroupV2(resp.SecurityGroup.Id); deleteErr != nil {
+			return nil, errors.Newf(deleteErr, "failed to roll back security group with id: %s", resp.SecurityGroup.Id)
```
wallyworld: Include the fact that we were creating tags, like the message below, otherwise there's no hint why the rollback was attempted

wallyworld (approved):
Thank you
We'll need to tag a new release v2.1.0 since this is a breaking api change

adisazhar123:
Doing last round of tests with Juju then I will proceed to merge
Edit:
I will keep this open for testing the integration in Juju side

PR #101, opened 2025-07-14
feat: add tags when updating a security group

adisazhar123:
This PR modifies an existing function, `UpdateSecurityGroupV2`, to receive a `tags` argument to attach tags to an existing security group. It also handles rolling back to the old security group if tag creation fails.

This is a follow up of: https://github.com/go-goose/goose/pull/100

### DIFF
--- a/neutron/neutron.go
+++ b/neutron/neutron.go
@@ -466,8 +466,8 @@ func (c *Client) DeleteSecurityGroupV2(groupId string) error {
 	return err
 }
 
-// ShowSecurityGroupV2 finds a security group by ID.
-func (c *Client) ShowSecurityGroupV2(groupId string) (*SecurityGroupV2, error) {
+// GetSecurityGroupV2 finds a security group by ID.
+func (c *Client) GetSecurityGroupV2(groupId string) (*SecurityGroupV2, error) {
 	var resp struct {
 		SecurityGroup SecurityGroupV2 `json:"security_group"`
 	}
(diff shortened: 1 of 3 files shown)


neutron/neutron.go
```diff
@@ -466,8 +466,23 @@ func (c *Client) DeleteSecurityGroupV2(groupId string) error {
 	return err
 }
 
+// ShowSecurityGroupV2 finds a security group by ID.
+func (c *Client) ShowSecurityGroupV2(groupId string) (*SecurityGroupV2, error) {
```
wallyworld: This is better named "GetSecurityGroupV2"

neutron/neutron.go
```diff
@@ -488,6 +503,41 @@ func (c *Client) UpdateSecurityGroupV2(groupId, name, description string) (*Secu
 	return &resp.SecurityGroup, nil
 }
 
+// UpdateSecurityGroupV2 updates the name, description, and tags (if specified) of the given group.
+// When tags are passed, it will overwrite the existing tags that are attached to the group.
+func (c *Client) UpdateSecurityGroupV2(groupId, name, description string, tags []string) (*SecurityGroupV2, error) {
+	existingGroup, err := c.ShowSecurityGroupV2(groupId)
+	if err != nil {
+		return nil, errors.Newf(err, "failed to update security group with ID: %s", groupId)
```
wallyworld: the error here isn't updating but getting the existing one

neutron/neutron.go
```diff
@@ -488,6 +503,41 @@ func (c *Client) UpdateSecurityGroupV2(groupId, name, description string) (*Secu
...
+	}
+	oldName := existingGroup.Name
+	oldDescription := existingGroup.Description
+
+	securityGroup, err := c.updateSecurityGroupV2(groupId, name, description)
+	if err != nil {
+		return nil, err
```
wallyworld: this is the place where we'd want to annotate the error
Also, error annotations start with a verb, ie "updating security group ..."

neutron/neutron.go
```diff
@@ -488,6 +503,41 @@ func (c *Client) UpdateSecurityGroupV2(groupId, name, description string) (*Secu
...
+	if len(tags) == 0 {
+		return securityGroup, nil
+	}
+
+	// Updates the tags for the group.
+	err = c.ReplaceAllTags("security-groups", securityGroup.Id, tags)
+	if err != nil {
+		// Rollback to old name and description if updating tags failed.
+		if _, updateErr := c.updateSecurityGroupV2(groupId, oldName, oldDescription); updateErr != nil {
+			return nil, errors.Newf(updateErr, "updating tags failed and attempt to roll back the security group failed. security group with id: %s", securityGroup.Id)
```
wallyworld: '... failed for security group with id: %s"`

wallyworld (approved):
Just a few small fixes.

neutron/neutron.go
adisazhar123: Done. Thanks!

neutron/neutron.go
adisazhar123: Addressed. Thanks.

neutron/neutron.go
adisazhar123: That sounds better.

neutron/neutron.go
adisazhar123: Done.

PR #105, opened 2026-04-30
feat: add v3 Keystone trust auth support

manadart:
This adds support for [Keystone Trusts](https://docs.openstack.org/keystone/2026.1/user/trusts.html).

The change is quite simple - a single, optional extra field in the `Credentials` struct for supplying a trust ID.

PR #106, opened 2026-05-21
fix: trust auth JSON tags

manadart:
Support for Keystone trusts was added under https://github.com/go-goose/goose/pull/105.

The original change was correct, but the struct tags were changed before landing. As it turns out, the specific form is required by O7k, so we restore it here.


SimonRichardson (approved):
That's not a nice tag, but it is what it is.

wallyworld (approved):

PR #107, opened 2026-06-29
feat: add cinder DeleteAttachment for stale volume attachments

xtrusia:
Adds `DeleteAttachment` to remove a Cinder volume attachment via the attachments API (microversion 3.27), to free a volume left wedged in `detaching` after its instance was deleted mid-detach.
Cinder refuses with 409 while the instance still uses the volume, so non-duplicate 409s now map to a new `Conflict` error (`errors.IsConflict`) — split into its own commit since it affects all goose APIs via `handleError`.

`go test ./...` passes; new tests cover the 409 mapping and the cinder path.

### DIFF
--- a/cinder/autogenerated_client.go
+++ b/cinder/autogenerated_client.go
@@ -405,10 +405,14 @@ type GetVolumesDetailParams struct {
 }
 
 type VolumeAttachment struct {
-	Device   string `json:"device"`
-	Id       string `json:"id"`
-	ServerId string `json:"server_id"`
-	VolumeId string `json:"volume_id"`
+	Device string `json:"device"`
+	Id     string `json:"id"`
+	// AttachmentId is the id of the attachment itself. On modern clouds Id is
+	// the volume id, so callers targeting a specific attachment (e.g. for
+	// os-force_detach) must use this field.
+	AttachmentId string `json:"attachment_id"`
+	ServerId     string `json:"server_id"`
+	VolumeId     string `json:"volume_id"`
 }
 
 type Volume struct {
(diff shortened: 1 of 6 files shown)


cinder/autogenerated_client.go
```diff
@@ -1699,3 +1703,24 @@ func listAvailabilityZones(client *Client) (*GetAvailabilityZonesResults, error)
...
+}
+
+// attachmentsMicroversion is the block-storage microversion at which the
+// attachments API (used by deleteAttachment) became available.
+const attachmentsMicroversion = "volume 3.27"
+
+// Deletes a volume attachment via the block-storage attachments API.
+func deleteAttachment(c *Client, args DeleteAttachmentParams) error {
+	requestData := goosehttp.RequestData{
+		ReqHeaders:     http.Header{"OpenStack-API-Version": []string{attachmentsMicroversion}},
```
wallyworld, line 1720: Should we use `X-OpenStack-Nova-API-Version` for maximum compatibility?

Microversions were introduced in compute api 2.1. Don't we need to check that if we are talking to api version 2.0 we need to return a not supported type error?

http/client.go
```diff
@@ -423,6 +423,9 @@ func handleError(URL string, resp *http.Response) error {
 		if dupExp.Match(errBytes) {
 			return errors.NewDuplicateValuef(httpError, "", string(errBytes))
 		}
+		if resp.StatusCode == http.StatusConflict {
+			return errors.NewConflictf(httpError, "", string(errBytes))
```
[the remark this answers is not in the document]
wallyworld: This seems like a candidate for a drive by fix across the board

cinder/autogenerated_client.go
xtrusia: > Should we use `X-OpenStack-Nova-API-Version` for maximum compatibility?
> 
> Microversions were introduced in compute api 2.1. Don't we need to check that if we are talking to api version 2.0 we need to return a not supported type error?

You might mean OpenStack-API-Version? as this is cinder. I added commit for the mention. ( and from 3.27)
https://docs.openstack.org/api-ref/block-storage/v3/#attachments-attachments

Also added NotImplemented for unsupported microversion
Thanks

wallyworld (approved):

PR #108, opened 2026-08-09
Preserve default transport settings for non-validating clients

xtrusia:
Non-validating clients now preserve proxy and connection settings from the default HTTP transport while still disabling TLS certificate verification.

Related issue: [juju/juju#20141](https://github.com/juju/juju/issues/20141)

##TEST##

Tested with a Juju 3.6.19 controller using Goose v5.1.4.
Direct access from the controller to Keystone was blocked, while access through the configured HTTP proxy returned HTTP 300.

Before this patch, the provider precheck bypassed the proxy:

```console
$ juju upgrade-controller -c "$CONTROLLER" \
    --dry-run --agent-version=3.6.20

ERROR cannot make API call to provider:
Get "http://<keystone>/": dial tcp <keystone>: connect: connection refused
```

Exit code: `1`

The same controller was then upgraded to an agent built with Goose v5.1.4 plus this patch.
Under the same proxy and firewall configuration, the same command succeeded:

```console
$ juju upgrade-controller -c "$CONTROLLER" \
    --dry-run --agent-version=3.6.20

best version:
    3.6.20
upgrade to this version by running
    juju upgrade-controller
```

Exit code: `0`

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -130,17 +130,24 @@ func WithInsecureHTTPClient(client *http.Client) Option {
 	}
 }
 
+func newInsecureHTTPClient() *http.Client {
+	transport, ok := http.DefaultTransport.(*http.Transport)
+	if ok {
+		transport = transport.Clone()
+	} else {
+		transport = &http.Transport{}
+	}
+	if transport.TLSClientConfig == nil {
+		transport.TLSClientConfig = &tls.Config{}
+	}
+	transport.TLSClientConfig.InsecureSkipVerify = true
+	return &http.Client{Transport: transport}
+}
+
 func newOptions() *options {
 	return &options{
 		httpHeadersFunc: goosehttp.DefaultHeaders,
 		httpClient:      &http.Client{},
-		insecureHTTPClient: &http.Client{
-			Transport: &http.Transport{
-				TLSClientConfig: &tls.Config{
-					InsecureSkipVerify: true,
-				},
-			},
-		},
 	}
 }
 
@@ -219,6 +226,9 @@ func NewNonValidatingPublicClient(baseURL string, logger logging.CompatLogger, o
 	for _, option := range options {
 		option(opts)
 	}
+	if opts.insecureHTTPClient == nil {
+		opts.insecureHTTPClient = newInsecureHTTPClient()
+	}
 
 	return &client{
 		baseURL: baseURL,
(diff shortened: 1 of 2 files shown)


client/client.go
```diff
@@ -219,6 +239,9 @@ func NewNonValidatingPublicClient(baseURL string, logger logging.CompatLogger, o
 	for _, option := range options {
 		option(opts)
 	}
+	if opts.insecureHTTPClient == nil {
```
wallyworld, line 242: `newOptions()` does this

```
func newOptions() *options {
	return &options{
		httpHeadersFunc: goosehttp.DefaultHeaders,
		httpClient:      &http.Client{},
		insecureHTTPClient: &http.Client{
			Transport: &http.Transport{
				TLSClientConfig: &tls.Config{
					InsecureSkipVerify: true,
				},
			},
		},
	}
}
```
So it already sets the insecureHTTPClient. And it does so incorrectly ignoring any env proxies.
So that's where the fix is needed.
ie that's where `newInsecureHTTPClient()` should be called.

client/client.go
```diff
@@ -130,17 +131,36 @@ func WithInsecureHTTPClient(client *http.Client) Option {
...
+				Timeout:   30 * time.Second,
+				KeepAlive: 30 * time.Second,
+			}).DialContext,
+			ForceAttemptHTTP2:     true,
+			MaxIdleConns:          100,
+			IdleConnTimeout:       90 * time.Second,
+			TLSHandshakeTimeout:   10 * time.Second,
+			ExpectContinueTimeout: 1 * time.Second,
+		}
+	}
```
wallyworld, line 152: You don't need any of this - the copilot suggestion seems a bit wrong.

```
    transport := http.DefaultTransport
    if transport.TLSClientConfig == nil {
		transport.TLSClientConfig = &tls.Config{}
	}
	transport.TLSClientConfig.InsecureSkipVerify = true
    return &http.Client{Transport: transport}
```

wallyworld (commented):
See my comments - hopefully they make sense.

wallyworld (approved):
Thank you

client/client.go
```diff
@@ -130,17 +130,25 @@ func WithInsecureHTTPClient(client *http.Client) Option {
...
-		httpClient:      &http.Client{},
-		insecureHTTPClient: &http.Client{
-			Transport: &http.Transport{
-				TLSClientConfig: &tls.Config{
-					InsecureSkipVerify: true,
-				},
-			},
-		},
+		httpHeadersFunc:    goosehttp.DefaultHeaders,
+		httpClient:         &http.Client{},
```
wallyworld: Actually, the httpClient should also preserve the transport proxy info etc

